PR #10, opened 2022-04-28
use prefix in pdf title

garyverhaegen-da:


docs/setup-sphinx-source-tree.sh
```diff
@@ -46,3 +46,6 @@ for file in pdf html; do
         sed -i "s|$var = u'.*'|$var = u'$prefix'|" $SPHINX_DIR/configs/$file/conf.py
     done
 done
+
+# Title page on the PDF
+sed -i "s|Version : .*|Version : $prefix|" $SPHINX_DIR/configs/pdf/conf.py
```
realvictorprm, line 51: sadly I'm no sed expert so what does this do?

S11001001 (approved):

docs/setup-sphinx-source-tree.sh
garyverhaegen-da: Let's deconstruct this one piece at a time:

- `-i` means "change in place". Without this flag, the default behaviour of sed is to take the file argument as input, and print the result on its stdout.
  ```
  sed -i "$CMD" "$FILE"
  ```
  is basically equivalent to
  ```
  sed "$CMD" "$FILE" > "$FILE".tmp && mv "$FILE.tmp" "$FILE"
  ```
- `"s|Version : .*|Version : $prefix|"` is the command given to `sed`. `sed` commands start with one letter, which determines the number of arguments; `s` is (by far) the most used command in my experience, and it takes three arguments. The first character after `s` is the argument separator; in most cases people choose to use `/`, but I deal with URLs pretty often so I have gotten in the habit of using `|` instead, which tends to have fewer collisions with the arguments themselves. Another popular choice is `:`. The `s` command `s`ubstitutes the first argument for the second argument; think of it as a filter that will be applied to each line in the input file.
- The first argument to `s`, `Version : .*`, is a regular expression. `Version : ` is composed of only "normal" characters and so matches itself. The last two characters are special in regex syntax: `.` matches any character, and `*` repeats the previous character as often as possible. So this matches any string that has `Version : ` from the `V` up until the end of the line.
- The second argument is `Version : $prefix`. Note that the entire `sed` command is in double quotes, so the `$prefix` part here is a simple Bash-level shell substitution, before `sed` sees the string at all. In our case, `$prefix` is set to the target release number for the docs website, extracted from the `LATEST` file. In this commit, it's `2.1.1`. So the actual argument `sed` sees is `Version : 2.1.1`.
- The third argument to `s` is empty, but we still need a separator to distinguish it from the second argument, which is why the `sed` command ends on a pipe (`|`).
- `$SPHINX_DIR/configs/pdf/conf.py` is the file we want to change.

So, concretely, in this case, what this does is look for this line in the `conf.py` file for the PDF generation:
```
            \small Version : 2.2.0-snapshot.20220412.9721.0.b7c4bdab
```
and replace it with this line:
```
            \small Version : 2.1.1
```
so we get the expected title in the final PDF.

PR #14, opened 2022-06-16
fix PDF ts links

garyverhaegen-da:
Reported by Ben on Slack: in the current PDF, the links are dead (because they are relative links). This PR changes the build script to turn them into absolute links in-between building the HTML and PDF versions.

This isn't "clean" or "robust" by any means, but I can't think of any other way to solve this.


docs/build.sh
```diff
@@ -26,14 +27,21 @@ mkdir -p $BUILD_DIR/source $BUILD_DIR/sphinx-target
...
+    sphinx-build -W --color -b $target -c $BUILD_DIR/source/configs/$name $BUILD_DIR/source/source $BUILD_DIR/sphinx-target/$name
+)
+
+build html html
+
+for ts_lib in types ledger react; do
+    sed -i \
+        "s|<|<https://docs.daml.com/$PREFIX/app-dev/bindings-ts/|" \
+        $BUILD_DIR/source/source/app-dev/bindings-ts/daml-$ts_lib.rst
 done
```
S11001001, line 43:
```suggestion
sed -i \
    "s|<|<https://docs.daml.com/$PREFIX/app-dev/bindings-ts/|" \
    $BUILD_DIR/source/source/app-dev/bindings-ts/daml-*.rst
```

LATEST
```diff
@@ -1,5 +1,5 @@
 {
-  "daml": "2.3.0-snapshot.20220509.9874.0.0798fe15",
-  "canton": "20220510",
+  "daml": "2.3.0-snapshot.20220614.10077.0.9c5c72a2",
+  "canton": "20220615",
```
S11001001, line 3: Does this go here?

S11001001 (approved):

LATEST
garyverhaegen-da: Yes? I'm not sure I follow the question. This is indeed the place where we put the Canton version.

If you're asking whether the version bump should be part of this PR, strictly speaking, no, it doesn't need to be. I just noticed it'd been a while since we bumped it and thought I might as well do it while I was there.

This will update the live https://docs.daml.com/2.3.0/ once merged.

akrmn:
I used a different approach in https://github.com/digital-asset/daml/pull/14225. Since we already have per-target `conf.py`, each with its `rst_prolog`, I thought we could define the typescript links there, with an absolute target in the `pdf` config. I think that's a bit cleaner since it keeps the logic in the daml repo.

This also fixes the links in [`docs/source/app-dev/bindings-ts/index.rst`](https://github.com/digital-asset/daml/blob/main/docs/source/app-dev/bindings-ts/index.rst), which iiuc this PR doesn't.

A drawback of my approach is that the fix would only land with the next release, but I hope that's okay given the other considerations.

garyverhaegen-da:
Thanks @akrmn !

PR #39, opened 2022-10-03
Add Daml Finance

garyverhaegen-da:
@carrielaben-da is there a better place to put this?

cc @brianweir-da


carrielaben-da (commented):
This is a good place for it to end up, but for now perhaps it should be in Early Access?

garyverhaegen-da:
Superseded by #40.

PR #40, opened 2022-10-06
Add Daml-Finance

brianweir-da:


garyverhaegen-da:
This now works for me as is.

@carrielaben-da since this would supplant #39, I assume your remark there still holds. However, there doesn't seem to be an "Early Access" top-level title anymore. Should we add one for this, or has it moved somewhere else?

brian-weir:
@carrielaben-da, the 

> > This now works for me as is.
> > @carrielaben-da since this would supplant #39, I assume your remark there still holds. However, there doesn't seem to be an "Early Access" top-level title anymore. Should we add one for this, or has it moved somewhere else?
> 
> There's an Early Access section in the TOC-test branch I'm currently working on, so it is being added back in anyway - not sure what the easiest way to handle these parallel changes is?

@carrielaben-da Probably the easiest way to handle this is that either you merge your changes to `main` and I update my branch with your changes. Or, since this branch is ready to be merged now that we merge it and you pull in the new changes to your branch and simply remove the `daml-finance` link that I've created in the `Create Daml Apps` section.

garyverhaegen-da (approved):

PR #41, opened 2022-10-08
artifactory

simonletort-da:
more details on how to use artifactory


garyverhaegen-da:
Hi @simonletort-da,

Unfortunately, this repo is not set up to accept external contributions. We also really do want an API key here, not an Identity Token.

PR #43, opened 2022-10-13
fix snapshot upload

garyverhaegen-da:
I wasn't quite sure how to pick the finance version; this seems adequate if we assume we always build snapshots reasonably on time.


matteolimberto-da:
> I wasn't quite sure how to pick the finance version; this seems adequate if we assume we always build snapshots reasonably on time.

I think this makes sense.

Ideally we would like to have the D.F. docs generated per-package, which can in principle be released independently. Given that we are not there yet, picking the latest version of Daml Finance is a reasonable approximation

matteolimberto-da (approved):

PR #59, opened 2022-11-14
enable toc ref

garyverhaegen-da:
In [daml#15449] we want to link to a TOC page, which doesn't exist there. So we resort to the same trick as for canton <-> daml links: introduce a phony reference on the daml side and reify it here.

[daml#15449]: https://github.com/digital-asset/daml/pull/15449


carrielaben-da (approved):
LGTM

PR #72, opened 2022-11-29
removing duplicate link to writing daml menu

katmurp:
Removing an actual duplicate menu item from the Reference section.


katmurp:
... reverting, because it's too confusing to figure out.

katmurp:
I wonder tho, I might retry this.

garyverhaegen-da:
As far as I can tell this is the only occurrence of `writing-daml` in the entire repo, so if there's duplication of something it's not here.

PR #73, opened 2022-11-29
removing duplicate link to writing daml menu

katmurp:
Let's give this a go @garyverhaegen-da I'm nervous about it but let's see what happens maybe.

### DIFF
--- a/docs/index/index.rst
+++ b/docs/index/index.rst
@@ -105,7 +105,6 @@ Daml Documentation
    :hidden:
    :caption: Reference
 
-   writing-daml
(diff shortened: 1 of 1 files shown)


garyverhaegen-da:
> ok, I don't understand the failure message

The secret to reading error messages is learning (through experience mostly) which parts you can ignore and which part (usually a tiny one) is actually useful. In this case the useful part of the error message is:
```
writing-daml.rst:document isn't included in any toctree
```
which makes sense to me given that this PR removes the only reference to `writing-daml`.

PR #74, opened 2022-11-29
Remove writing-daml from reference

jwhyte-da:
This content is already in "Write smart contracts with daml" section in Create Daml Apps


katmurp:
You can't remove this, I tried earlier. Removing it breaks the build. You have to remove the page in Daml docs also but we don't know what unusual effects that might have.

PR #71, opened 2022-11-28
Added custom-views doc to early access TOC.

ray-roestenburg-da:
Reflects https://github.com/digital-asset/daml/pull/15721; do not merge until

* [ ] that PR is merged
* [ ] `LATEST` here has been updated with a `daml` snapshot that includes that PR
* [ ] rerun CI, which will fail until the above are satisfied

PR #69, opened 2022-11-28
Update latest

brian-weir:

PR #75, opened 2022-11-30
bump deps

garyverhaegen-da:


garyverhaegen-da:
This supersedes both #69 and #71 because it's difficult to do them separately.

This should reduce the duplicated toc issue in the Daml language tutorial, but not solve it entirely (specifically, this includes the removal of [the toc in intro_0](https://github.com/digital-asset/daml/pull/15718), but **not** the removal of [the toc in writing-daml](https://github.com/digital-asset/daml/pull/15723), which will have to wait for tomorrow).

brian-weir (approved):

PR #76, opened 2022-11-30
Added common metrics to table of contents

miklos-da:


garyverhaegen-da:
@miklos-da Is it intended that this branch has no change at all?

miklos-da:
> @miklos-da Is it intended that this branch has no change at all?

It seems @ray-roestenburg-da has added the TOC entry already in #79 — closing this.

PR #78, opened 2022-12-01
Remove Developer Tools

carrielaben-da:
Removes Developer Tools from the Early Access menu

[CHANGELOG_BEGIN]
[CHANGELOG_END]

### DIFF
--- a/LATEST
+++ b/LATEST
@@ -1,5 +1,5 @@
 {
-  "daml": "2.5.0-snapshot.20221129.11056.0.4a554ac2",
+  "daml": "2.6.0-snapshot.20221201.11069.0.1cb6bc63",
   "canton": "20221129",
   "daml_finance": "1.0.0",
   "prefix": "2.5.0"
--- a/docs/index/index.rst
+++ b/docs/index/index.rst
@@ -134,7 +134,6 @@ Daml Documentation
 
    concepts/interoperability
    operating-daml
-   tools
    tools/export/index
    tools/non-repudiation
    tools/visual


garyverhaegen-da:
We'll have to decide here whether we keep the `daml/reference/index.rst` file at all, but for this PR I'd still keep it, either in its rightful place if we want to keep it long-term, or in a section `TRASH` at the end of the toc, since removing it will require another daml snapshot roundtrip.

katmurp:
I have some questions. Why is the snapshot 2.6? Also, we still appear to be reference subpages from tools/ i.e. ```concepts/interoperability
   operating-daml
   tools
   tools/export/index
   tools/non-repudiation
   tools/visual```
Won't removing the top menu item `tools` mean any sub items will break?

carrielaben-da:
> > I have some questions. Why is the Daml snapshot 2.6?
> 
> Because we have started the RC process, and code changes no longer go into the 2.5.0 release. We could be as picky with docs changes and stick to the 2.5.0 branch, if we make sure to backport all relevant docs changes in the daml repo. Or we can use a 2.6 snapshot if we're confident all of the docs changes since the 2.5.0 cutoff are "valid" for the 2.5.0 codebase.
> 
> My preference would be to stick to 2.5.0 snapshots here, but it does add a substantial burden to any docs fix on the daml repo, so we may be better served by 2.6.0 snapshots for now.
> 
> > Also, we still appear to be referencing subpages from tools/ i.e.
> > ```
> >    concepts/interoperability
> >    operating-daml
> >    tools
> >    tools/export/index
> >    tools/non-repudiation
> >    tools/visual
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Won't removing the top menu item `tools` mean any sub items will break?
> 
> Concretely, we've removed the `tools.rst` file, but we've kept the `tools` folder and all the files in it. The `tools.rst` page was basically just a toc, so likely responsible for some of the duplication bugs we've seen on the 2.5.0 website.
> 
> > I wouldn't call something 'Trash' in a public-facing doc but a lighter version of the term might work.
> 
> I don't consider the `/2.5.0` website public, but "To be deleted" works just as well.

Just to be clear, what will appear on the website once 2.5 is released?

garyverhaegen-da:
> Just to be clear, what will appear on the website once 2.5 is released?

Here's the current process. At the point in time where we declare the release stable (by unticking the "pre-release" marker on the yet-to-be-created `2.5.0` GitHub release), the files under `https://docs.daml.com/2.5.0` will be copied to the root website (`https://docs.daml.com`). So the short answer is "whatever is under `/2.5.0` when we make the release. And that is controlled by the `LATEST` file in this repo.

It doesn't matter what the daml, canton, or finance versions are called. Whatever we pick in `LATEST` is put under `/2.5.0` (as long as the `prefix` entry of `LATEST` says `2.5.0`), and that's all that matters.

garyverhaegen-da:
Subsumed by #80.

PR #80, opened 2022-12-06
Toc poc

garyverhaegen-da:


docs/index/_toc.yml
```diff
@@ -0,0 +1,964 @@
+root: index
```
matteolimberto-da, line 1: Is this file auto-generated somehow?

It seems that changing any file in the Daml Finance doc would then require us to modify this file. Could you confirm if my understanding is correct? 

cc @brianweir-da

garyverhaegen-da: Changing existing files, no; adding or removing files, yes. That's the same for all other dependencies.

PR #92, opened 2022-12-15
adding backport instructions

katmurp:
Full instructions on how to deal with situations in which you need to make a change to an earlier version.


garyverhaegen-da (approved):
Thanks!

PR #107, opened 2023-01-13
update LATEST

da-katmurp:
Wasn't seeing any changes come in from the last couple of weeks. Realized we have to manually update this and not rely on someone else doing it. Wondering how this could be automated?


da-katmurp:
The build is breaking due to a new page in Daml that has no TOC entry. How can we deal with this?

da-katmurp:
OK, I'll just add an entry...although this means that, even though we have no idea if the doc should be included in the next release or not, it will be regardless of it's state, or mean another heap of kerfuffles at release time.

garyverhaegen-da:
> OK, I'll just add an entry...although this means that, even though we have no idea if the doc should be included in the next release or not, it will be regardless of its state, and maybe mean another heap of kerfuffles at release time.
> 
> We are also just guessing where it should go to fix the build error. This is asking for trouble really.

So, in principle, this is not a random, unknown page, this is the result of a peer-reviewed pull request on the daml repo. If it's on the main branch of the daml repo (which snapshots are taken from), it means that:
- It has been reviewed and approved, and
- It is meant to be included in the public docs of the next version.

garyverhaegen-da (approved):

da-katmurp:
This is an assumption that we currently have no visibility on, as this doc is still in draft even though a chunk of it has been pushed to main and there are outstanding reviews going on. 

So anything can happen. And we are guessing on a toc with no context. This is one way that docs become chaotic.

PR #116, opened 2023-01-19
update LATEST

da-katmurp:
Updating LATEST, mainly for this one: https://trello.com/c/5ynRRuCD (which, for info, I would push to live immediately as ignoring complaints is really poor customer service) ... but also to see what happens with canton docs now they just released an "earlier" version of their docs, i.e. 2.5.1 not 2.6.0.


gerolf-da (approved):

gerolf-da:
Since the changes for 2.5.1 are now reverted on main, I think everything should be back on track for making progress on 2.6.x.

PR #121, opened 2023-01-23
clean up download script

garyverhaegen-da:
Reducing duplication in preparation for future improvements.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -80,7 +80,6 @@ jobs:
                echo "Pushhing to $prefix is not currently supported."
                exit 1
              fi
-             bin/download
              bin/build
              mv workdir/target/html-docs-*.tar.gz /tmp/workspace/
            EOF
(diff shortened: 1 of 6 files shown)


garyverhaegen-da:
Small improvement here: if the downloads are already there, we don't re-download them.

da-katmurp (approved):
So we won't have to run download anymore at all?

garyverhaegen-da:
Actually added a slightly bigger improvement: `build` and `live-preview` now automatically call `download` (which is "safe" as it no longer overwrites existing tarballs) so there should no longer be a use-case for calling `download` explicitly. Updated the README & CI scripts to match.

garyverhaegen-da:
> So we won't have to run download anymore at all?

Exactly. It's not a huge thing, but one fewer step seems good.

da-katmurp:
any chance of updating LATEST at the same time?

garyverhaegen-da:
> @garyverhaegen-da any chance of updating LATEST at the same time?

Same time as what?

da-katmurp:
> > @garyverhaegen-da any chance of updating LATEST at the same time?
> 
> Same time as what?

A download would seem like the best moment to update LATEST.

PR #122, opened 2023-01-23
bump

garyverhaegen-da:


da-katmurp (approved):
Is it possible to auto-run this at the same time the tarballs come in maybe?

PR #126, opened 2023-01-24
moving resource mgmt to correct spot

da-katmurp:
I had put the Resource Management docs in the wrong place previously.

This PR corrects that.


_toc.yml
```diff
@@ -40,7 +39,7 @@ subtrees:
   - file: daml/considerations
     title: "Introduction"
   - file: create-daml-apps/intro/index
-    title: "Write Smart Contracts with Daml"
+    title: "Write Smart Contracts with Damls"
```
garyverhaegen-da, line 42: You probably want to remove this line

garyverhaegen-da (approved):
Thanks!

_toc.yml
da-katmurp: well spotted

PR #128, opened 2023-01-24
add mention of search reindexing

loiskilberg-da:
Add a section mentioning when and how to reindex the site search, such as when a new version of the docs is released.

[CHANGELOG_BEGIN]
[CHANGELOG_END]


garyverhaegen-da (approved):
Thanks!

PR #155, opened 2023-02-08
build release 2.5.3

garyverhaegen-da:
As for 2.5.2, this is the result of copying the 2.5.2 folder and changing LATEST to use 2.5.3 instead.


gerolf-da (approved):

gerolf-da:
Do we have a need to keep docs for patch releases that are not the latest patch release around? Could we just stick with a folder for 2.5.x and continually keep bumping the version?

garyverhaegen-da:
> Do we have a need to keep docs for patch releases that are not the latest patch release around? Could we just stick with a folder for 2.5.x and continually keep bumping the version?

There's a good argument for deleting 2.5.1 at this point, since we anyway can't do patches to it on either the canton or daml sides. We do need both 2.5.2 and 2.5.3 right now, though, because 2.5.2 is root and 2.5.3 shouldn't be until a bit later.

PR #160, opened 2023-02-13
reduce dropdown size

garyverhaegen-da:
I'd really like an approval from @digital-asset/daml-product on this one.

When merged, this PR will reduce the size of the version dropdown on [docs.daml.com](https://docs.daml.com) to just the latest patch for each 2.x minor and a single 1.x version (1.18.3).

This will not delete any page, i.e. the 1.17.0 and 2.2.0 versions will still be available on, respectively, https://docs.daml.com/1.17.0 and https://docs.daml.com/2.2.0. They just won't appear in the dropdown anymore.

I'd really like to prune this more, but this seemed like a safe first step as I don't expect anyone to object. (but we'll see)


da-katmurp:
One question please: is to do with the menu not scrolling in an attempt to make it shorter?

garyverhaegen-da:
> One question please: is this to do with the menu not scrolling in an attempt to make it shorter?

Not really; the lack of scrolling _is_ annoying, but:
- we can only conveniently update the latest patch of each minor in both code and docs.
- our support policy is that we only support the latest patch for each minor.
- there's very little value in keeping older patches as the differences in docs are typically marginal and always strict improvements (we don't go through the effort of changing docs for an older version unless there's a strong need).
- at this point we really don't want anyone on 1.x, especially other than 1.18.3, and possibly also on 2.2.x or lower, so it's not clear who those additional versions of the docs would be useful for and we may not want to advertize that they are available.

da-katmurp:
> > One question please: is this to do with the menu not scrolling in an attempt to make it shorter?
> 
> Not really; the lack of scrolling _is_ annoying, but:
> 
> * we can only conveniently update the latest patch of each minor in both code and docs.
> * our support policy is that we only support the latest patch for each minor.
> * there's very little value in keeping older patches as the differences in docs are typically marginal and always strict improvements (we don't go through the effort of changing docs for an older version unless there's a strong need).
> * at this point we really don't want anyone on 1.x, especially other than 1.18.3, and possibly also on 2.2.x or lower, so it's not clear who those additional versions of the docs would be useful for and we may not want to advertize that they are available.

Cool, makes sense. We should fix the scroll issue too if we can.

gerolf-da (approved):

gerolf-da:
Approving on behalf of @bame-da, who verbally agreed to the change.

garyverhaegen-da:
> Approving on behalf of @bame-da, who verbally agreed to the change.

Thanks!

PR #162, opened 2023-02-13
Add intro to interfaces to ToCs

akrmn:
I missed adding this earlier, apologies. Is this enough to get this section to show up? if not please let me know so I can do the rest.

### DIFF
--- a/docs/2.5.1/docs/index/_toc.yml
+++ b/docs/2.5.1/docs/index/_toc.yml
@@ -61,6 +61,8 @@ subtrees:
         title: "Parties and Authority"
       - file: daml/intro/7_Composing
         title: "Composing Choices"
+      - file: daml/intro/13_Interfaces
+        title: "Interfaces"
       - file: daml/intro/8_Exceptions
         title: "Exception Handling"
       - file: daml/intro/9_Dependencies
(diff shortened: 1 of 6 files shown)


garyverhaegen-da:
Hi @akrmn,

I don't think there's anything to do here, actually. The interfaces intro is already part of the 2.6 ToC, and the file doesn't exist in the 2.5 docs (which makes sense: 2.5 was cut mid-December and the file was added mid-January).

If you're asking about how to add the page to 2.5 docs, that's a separate discussion. We'd first need to decide whether we want to do that; we generally don't add pages to past docs, but since 2.5.3 is the current root that may be warranted in this case.

I have no idea who can/should approve that kind of change.

On a technical level, if we wanted to add that page:

1. The process would have to start with a backport of the page to the `release/2.5.x` branch on the daml repo.
2. Once that backport is merged and we have a "split snapshot" (which atm I have to trigger manually for release branches), we can reference that snapshot in the `LATEST` (2.5) or `versions.json` (2.6) file, and do the appropriate ToC change.
3. We would probably want to only add it to the 2.5.3 branch, I think.

garyverhaegen-da:
Oh and note that [the 2.6 page is accessible](https://docs.daml.com/2.6.0/daml/intro/13_Interfaces.html), if not easy to find through links from the homepage.

akrmn:
@garyverhaegen-da thanks for explaining. To be honest that sounds like a lot of work, especially considering the 2.6.0 version is accessible and that it will be released before end of month, so I'm closing this PR.

garyverhaegen-da:
> To be honest that sounds like a lot of work

It is. We're looking into ways to lessen that in the future, but they will likely only be applied moving forward so won't be helpful for this specific case.

PR #163, opened 2023-02-15
update Canton requirements

garyverhaegen-da:
We want to update the root website without making a Canton release, and AFAICT the only way to do that is to move this page here.

We also want to move pages here generally speaking, so, bonus.


bame-da (approved):
ty. Are the corresponding pages gone from the Canton repo?

gerolf-da (approved):
thanks 👍

garyverhaegen-da:
> ty. Are the corresponding pages gone from the Canton repo?

Not yet but I've opened a PR to make that happen.

PR #164, opened 2023-02-15
remove unsupported 2.5.2 and 2.5.1

garyverhaegen-da:


garyverhaegen-da:
Now that they're not in the dropdown anymore (and given the difficulty of keeping them updated), I think we can drop these two versions.

carrielaben-da (approved):
Makes sense to me.

gerolf-da (approved):

PR #175, opened 2023-02-25
add contention page

loiskilberg-da:
Add a page describing how to design to avoid contention. It will go under the resource management section.

[CHANGELOG_BEGIN]
[CHANGELOG_END]

### DIFF
--- a/docs/2.6.0/_toc.yml
+++ b/docs/2.6.0/_toc.yml
@@ -162,6 +162,7 @@ subtrees:
           titlesonly: True
           entries:
           - file: daml/resource-management/latency-and-throughput
+          - file: daml/resource-management/contention
           - file: daml/resource-management/active-contract-set
       - file: app-dev/ledger-api
         title: "Build Integration with the Ledger API"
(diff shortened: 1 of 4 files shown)


docs/2.6.0/_toc.yml
```diff
@@ -162,6 +162,7 @@ subtrees:
           titlesonly: True
           entries:
           - file: daml/resource-management/latency-and-throughput
+          - file: daml/resource-management/contention
```
loiskilberg-da, line 165: @da-katmurp Does this look like a good place in the TOC?

da-katmurp (changes requested):
This was extremely hard going. 

I made a number of suggestions that will help the reader cope with so many words.

The language was OK until I saw Bernhard's way of writing creeping in here and there, and at those points I edited language to remove passive, random modals, future tense, random capital letters. At that point, it became impossible to maintain the narrative in my mind, and do the proofing at the same time, so I dropped trying to figure out what it all means.

I admit, I was weary at the last section so did not do too thorough a job of that. However, I believe a lot of that last section is in the wrong place anyway. Happy to review again once it's easier to navigate.

loiskilberg-da:
Closing this PR in favor of https://github.com/digital-asset/docs.daml.com/pull/319

PR #185, opened 2023-03-02
prepare 2.5.4

garyverhaegen-da:
We are currently finalizing the 2.5.4 release, so we should have a subsite for it, but it's not quite final yet so we're not updating the root or dropdiwn quite yet.


garyverhaegen-da:
This is just a straight up copy of `docs/2.5.3` to `docs/2.5.4`, with a goal of removing `docs/2.5.3` when we update the root & dropdown, hopefully soon (as soon as the Canton team has collated the release notes for 2.5.4).

loiskilberg-da (approved):

PR #197, opened 2023-03-08
Requires is now enabled by default

basvangijzel-DA:
Requires is now enabled by default so the warning for 1.dev does not apply any more.

### DIFF
--- a/docs/2.6.0/docs/daml/reference/interfaces.rst
+++ b/docs/2.6.0/docs/daml/reference/interfaces.rst
@@ -359,11 +359,6 @@ Interface Functions
 Required Interfaces
 *******************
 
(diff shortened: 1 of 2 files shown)


garyverhaegen-da (changes requested):
Please mirror this change to the 2.7 folder, or confirm this really is a 2.6-only change.

akrmn (commented):
thanks @basvangijzel-DA!

garyverhaegen-da (approved):
Given the nature of the change and the context I have, I am assuming that this should indeed be mirrored to 2.7 and have taken the liberty to do so myself for expediency.

Please reach out to me if that was a mistake.

PR #198, opened 2023-03-08
fix gpg mentions

garyverhaegen-da:
Back in January, we updated out GPG key. At the time, in order to have all of our docs display the new key, I ran a script to directly update the corresponding HTML files on the S3 bucket, and updated the main branch of the [daml] repo (where the corresponding file was at the time).

This seemed sensible at the time as we did not have the ability to update old versions of the docs.

[daml]: https://github.com/digital-asset/daml

But now we do, and we have pushed new versions of old docs for 2.3.8, 2.3.9, 2.5.1, 2.5.2, 2.5.3, 2.5.4, and 2.4.3. And none of those had the GPG key change, since that had been done on the HTML files directly, and not on the source files on those old branches.

This is now rectified (for [2.3.x], [2.4.x], [2.5.x]; earlier branches are still a work in progress as building old branches can be tricky), but that does mean we need to republish the docs for the affected versions.

[2.3.x]: https://github.com/digital-asset/daml/pull/16463
[2.4.x]: https://github.com/digital-asset/daml/pull/16464
[2.5.x]: https://github.com/digital-asset/daml/pull/16465

The docs for 2.6 and above were modified at the source back in January and thus are not affected by this issue.

Due to the large number of files, this PR may be hard to review, so here is a summary of what it contains:

- The `2.3.8` folder has been created as a copy of `2.3.9`. The `LATEST` file in `2.3.8` has been modified to reflect the `2.3.8` docs, except for the `daml` entry which points to `2.3.10` (the first one with the fixed GPG key).
- The `2.3.9/LATEST` file has similarly been updated for its `daml` entry to point to `2.3.10`.
- The `2.5.1` and `2.5.2` folders have been created as copies of the `2.5.3` folder, and their `LATEST` files have been adapted as above.

In short, to review this PR you should only review the various `LATEST` files that have been edited or added, and check that:

1. Their `prefix` entry matches their folder.
2. Their `daml` entry contains the GPG fix; `2.3.10` for the `2.3` releases, `2.4.3-snapshot.20230307.10604.0.9ad498d7` for the `2.4` release, and `2.5.4-snapshot.20230307.11092.0.e86e6a45` for the `2.5` releases.


gerolf-da (approved):
@garyverhaegen-da: do we have the version string appearing the documentation somehwere, so that it would now say `2.5.4-SNAPSHOT...` instead of `2.5.3`?

garyverhaegen-da:
> @garyverhaegen-da: do we have the version string appearing the documentation somehwere, so that it would now say `2.5.4-SNAPSHOT...` instead of `2.5.3`?

Not to my knowledge; as far as I know `prefix` is the only string from `LATEST` that gets into the docs, which is why it's important that it matches the folder name. (From 2.6.0 onward we take that from the folder name directly so there's no risk of desync.)

PR #200, opened 2023-03-08
Update Daml Finance to 1.1.5

matteolimberto-da:
This commit reverts the previous changes to the `toc` to avoid breaking URLs.

The 1.1.5 assembly also includes a hoogle index for Daml Finance


garyverhaegen-da (approved):
Thanks!

PR #222, opened 2023-03-30
bump canton & merge error codes

garyverhaegen-da:
This gets the docs build working following DACH-NY/canton#10858. We cannot have two error pages with the same content as they include labels; the two pages seemed similar enough that merging them seemed appropriate. I'm unclear on where the merged page should end up in the toc, though.


oliverse-da (approved):

carrielaben-da:
I was initially thinking that the final merged page should go in Reference but Help (as suggested by Bernhard) probably makes more sense. Within Help, at the top of the section above Troubleshooting would be my proposal.

garyverhaegen-da:
> I was initially thinking that the final merged page should go in Reference but Help (as suggested by Bernhard) probably makes more sense. Within Help, at the top of the section above Troubleshooting would be my proposal.

Moved to Help. I'll go ahead and merge this for now, feel free to move it around in a subsequent PR as needed. I think @yves-da will also want to re-bump tomorrow or next week to fix the formatting.

PR #244, opened 2023-04-19
Describe Observe vs ObserveAt

markus-da:
Fixes https://github.com/digital-asset/daml-finance/issues/797

### DIFF
--- a/docs/2.7.0/docs/daml-finance/concepts/contingent-claims.rst
+++ b/docs/2.7.0/docs/daml-finance/concepts/contingent-claims.rst
@@ -206,10 +206,11 @@ The Observation Parameter
 be observed at any given time (such as the “EURUSD” exchange rate in the example above). There are
 two ways of creating an observation in the library:
 
-- :ref:`Observe <constr-contingentclaims-core-observation-observe-30391>` is the original Claims
-  node described in the paper. It does not take a time parameter. Instead, you need to put a
-  ``when`` node in front of it. For simple claims this is quite straightforward, but for more
-  complex payoffs it can be difficult to correctly construct the claim.
+- :ref:`Observe <constr-contingentclaims-core-observation-observe-30391>` does not take a time
+  parameter. Instead, it is evaluated at the contract's acquisition time. You can put a ``When``
+  node in front of it, in order to fix the acquisition time. For simple claims this is quite
+  straightforward, but for more complex payoffs it can be difficult to correctly construct the
+  claim.
 - :ref:`ObserveAt <constr-contingentclaims-core-observation-observeat-8418>` is similar, but it
   takes a time parameter in addition. This makes it significantly easier to use, especially for
   complex expressions combining several features (e.g. multiple fixing dates, FX adjusted notional,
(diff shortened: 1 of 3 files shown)


brian-weir (approved):

docs/2.7.0/docs/daml-finance/concepts/contingent-claims.rst
```diff
@@ -203,7 +203,19 @@ The Observation Parameter
 =========================
 
 ``o`` is the type used to represent ``Observation``\ s, which are time-dependent quantities that can
-be observed at any given time (such as the “EURUSD” exchange rate in the example above).
+be observed at any given time (such as the “EURUSD” exchange rate in the example above). There are
```
matteolimberto-da, line 206: This section is meant to describe the type parameters involved in a claim, so the description of the specific observation combinators should probably live in a separate place.

There are also other combinators in addition to `Observe` and `ObserveAt`, such as `Const` and the algebraic combinators `+` `*`

docs/2.7.0/docs/daml-finance/concepts/contingent-claims.rst
```diff
@@ -203,7 +203,19 @@ The Observation Parameter
...
+two ways of creating an observation in the library:
+
+- :ref:`Observe <constr-contingentclaims-core-observation-observe-30391>` is the original Claims
```
matteolimberto-da, line 209: The paper doesn't really tell you anything about how to model an observation, so I wouldn't say it is the original operator :)

docs/2.7.0/docs/daml-finance/concepts/contingent-claims.rst
```diff
@@ -203,7 +203,19 @@ The Observation Parameter
...
+  node described in the paper. It does not take a time parameter. Instead, you need to put a
```
matteolimberto-da, line 210: You don't really need a `When` node in front of it. 

The observable is simply evaluated at the contract's acquisition time. 

The `When` is a way to fix the acquisition time, but in general an `Anytime` node also fixes an acquisition time

docs/2.7.0/docs/daml-finance/concepts/contingent-claims.rst
```diff
@@ -203,7 +203,19 @@ The Observation Parameter
...
+  ``when`` node in front of it. For simple claims this is quite straightforward, but for more
+  complex payoffs it can be difficult to correctly construct the claim.
+- :ref:`ObserveAt <constr-contingentclaims-core-observation-observeat-8418>` is similar, but it
+  takes a time parameter in addition. This makes it significantly easier to use, especially for
+  complex expressions combining several features (e.g. multiple fixing dates, FX adjusted notional,
+  amortizing notional, etc). It is also safer from a lifecycling point of view, since the time
```
matteolimberto-da, line 216: Luciano wouldn't agree with this statement, arguably `ObserveAt` is less safe :)

matteolimberto-da (changes requested):
I think the explanation is in the wrong place and not very clear from the perspective of a new user. 

We shouldn't say "use this, not that" but provide examples to let the user understand the value of each combinator. 

Happy to discuss how we can make this more user-friendly

docs/2.7.0/docs/daml-finance/concepts/contingent-claims.rst
markus-da: > so the description of the specific observation combinators should probably live in a separate place.

Where do think this belongs? In one of the existing places or in a new one?

matteolimberto-da:
Supplanted by https://github.com/digital-asset/docs.daml.com/pull/397

PR #252, opened 2023-05-02
delete old versions

garyverhaegen-da:
Should have done this a while ago.


carrielaben-da (approved):
Don't see any reason not to.

PR #259, opened 2023-05-04
recover ability to publish 2.1 docs

garyverhaegen-da:
This will rebuild them exactly as they are in the published version (note the 2.2 snapshot used).


garyverhaegen-da:
Actually that's not something we want as that would overwrite the GPG key change from January.

PR #275, opened 2023-05-08
Inifinite loop of evaluation in template-level let

matevarga-da:
This is an attempt to at least document the evaluation issue with template-level lets, as discussed in https://github.com/digital-asset/daml/issues/13470 and support ticket #00004485.

PR #277, opened 2023-05-11
New lines required around the list in rst.

matevarga-da:
The list is broken down into flowing text as I did not add new lines around it in my [previous pull](https://github.com/digital-asset/docs.daml.com/pull/275) request, sorry.


akrmn (approved):
thanks @matevarga-da!

PR #279, opened 2023-05-16
fix 2 links

garyverhaegen-da:
Unfortunately there's no label for the second link to point to on the Canton side, so I've made it work for now but it's likely to break again in the future.

(Or, rather, should it break in the future it will be a silent failure.)


carrielaben-da (approved):
Thanks!

PR #281, opened 2023-05-17
Spell check everything else

leonidr-da:
I'm sorry that my end-of-line white spaces trimmer is adding all that noise; I hope that you'll forgive me. To spell check, I 
```
$ for f in $(find . -type f -name "*.rst" ); do echo $f ; aspell list < $f ; done | sort | uniq -c > wrong.txt
```

and then manually scanned the result and fixed it in vscode. 99% of the result are false positives, but it took me about 20 minutes to scan through the entire list.

### DIFF
--- a/docs/2.5.5/docs/canton/architecture/requirements/requirements.rst
+++ b/docs/2.5.5/docs/canton/architecture/requirements/requirements.rst
@@ -1,7 +1,7 @@
 ..
      Copyright (c) 2023 Digital Asset (Switzerland) GmbH and/or its affiliates
 ..
-    
+
 ..
      Proprietary code. All rights reserved.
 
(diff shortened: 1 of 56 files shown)


docs/2.5.5/docs/canton/architecture/requirements/requirements.rst
```diff
@@ -1,7 +1,7 @@
 ..
      Copyright (c) 2023 Digital Asset (Switzerland) GmbH and/or its affiliates
 ..
-    
+
```
garyverhaegen-da, line 4: Fun fact: the Canton build actually fails when you make that change in the Canton repo.

garyverhaegen-da (approved):
Thanks!

PR #282, opened 2023-05-17
Replace the Developer Workflow image with an updated version

carrielaben-da:
Old version had all-caps DAML
Also a slight grammar edit


loiskilberg-da (approved):
looks good!

PR #284, opened 2023-05-22
Pruning: Clarifying a pruning edge-case statement for best practices

oliverse-da:


oliverse-da:
The completions and transaction stream queries actually end up using the overall start offset rather than the one of each batch, so the documentation is adquate without any changes.

loiskilberg-da (approved):

daravep (approved):
LGTM

PR #286, opened 2023-05-22
Bump canton version for 2.7.0 and adjust includes

yves-da:
Updates the canton version to the latest snapshot for 2.7.0; which also requires to adjust the `literalinclude` paths and the Canton Console Command reference (changed file location).

Depends on canton PRs #13003, #13042


yves-da:
Build fails as it depends on an updated canton documentation artefact.

garyverhaegen-da (approved):
Thanks!

docs/2.7.0/versions.json
```diff
@@ -1,5 +1,5 @@
 {
   "daml": "2.7.0-snapshot.20230406.11654.0.15ee6823",
-  "canton": "20230407",
+  "canton": "2.7.0-snapshot.20230522.10382.0.v967c73b6",
```
yves-da, line 3: @garyverhaegen-da 

Since yesterday, the Canton team uses the same versioning pattern as Daml.

As a consequence, I noticed that `/bin/bump` now won't find the most recent version anymore.

I looked at it quickly, but didn't have a quick fix for it. Probably `sort` needs to be used in some fashion in
```
canton=$($DIR/deps list canton | tail -1)
```
in the `bump` script.

garyverhaegen-da: Thanks, I'll take a look. Shouldn't block this PR though.

yves-da:
**DO NOT MERGE** – May have to fix the include for the Canton error codes as this has also changed.

PR #297, opened 2023-05-24
Drop documentation about Daml Ledger export

remyhaemmerle-da:
It is an alpha feature that is deprecated and does not support canton properly.


remyhaemmerle-da:
> Should you then also remove the actual page for it?

3792262e1acd9712baa37cb2103c95093c05b3e8

garyverhaegen-da:
For the build to work you'll need to drop it from `_toc.yml` too.

loiskilberg-da (approved):

PR #299, opened 2023-05-24
fix canton install instructions

garyverhaegen-da:
Note: we cannot fix 2.3.12 with the current infrastructure as it uses a `snippet` directive of the Canton preprocessor.


yves-da (approved):
LGTM.

The fix regarding `domain1.conf` is contained.

To be consistent with previous changes you'd have to go through with removing these files from the respective canton branches.

PR #310, opened 2023-06-02
Lift preview flag from grpc health check in 2.7.0

thibault-da:


yves-da (approved):
LGTM.

After the merge, this change should be published to the docs.daml.com page after the `main` branch has been built. You can then check your change, by selecting the `2.6.4` version (to get the URL with that version), and change it manually to `2.7.0`.

Also, since we have the docs.daml.com repo as a Git submodule in the canton repo; it may be necessary to update that reference (the commit hash). So that would be something to check after the merging this PR (and may require another PR on the canton repository).

PR #316, opened 2023-06-12
add a ping before downloading the ACS

andreaslochbihler-da:
This increases the likelihood of the timestamp being clean at the download

Part of https://github.com/DACH-NY/canton/issues/13368


carrielaben-da (commented):
LGTM but should definitely be approved by Yves.

yves-da (approved):
LGTM 👍

PR #318, opened 2023-06-12
fix dev dropdown

garyverhaegen-da:
The dropdown defaults to the first entry on the root site, so I'm putting the dev site as the second entry.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -151,7 +151,7 @@ jobs:
              --run bash \<<'EOF'
              set -euo pipefail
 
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -151,7 +151,7 @@ jobs:
              --run bash \<<'EOF'
              set -euo pipefail
 
-             DIR="$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
```
garyverhaegen-da, line 154: I'm not sure I understand how that happens, but apparently `BASH_SOURCE` is not defined on CircleCI.

loiskilberg-da (approved):

PR #330, opened 2023-06-20
Replace canton.io references in ethereum.rst

yves-da:
Instead of using RST cross-links hyperlinks to the documentation published on canton.io have been used.

As the documentation on canton.io is discontinued with the launch of the Canton Network these references have to be replaced.

This commit also uses an updated canton tarball (`versions.json) which uses Sphinx relative paths for the scaladoc links.

This should make it possible to move the Canton Console Commands and Error Codes reference pages to different directory locations without breaking the therein contained scaladoc links.


garyverhaegen-da:
What do you call relative links in this context? Specifically, will they work under version folders?

yves-da:
> What do you call relative links in this context? Specifically, will they work under version folders?

Instead of using this link (example):
`* **Scaladocs**: `UNSUPPORTED_OPERATION <../../canton/scaladoc/com/digitalasset/canton/ledger/error/CommonErrors$$UnsupportedOperation$.html>`_`
use this
`* **Scaladocs**: `UNSUPPORTED_OPERATION </canton/scaladoc/com/digitalasset/canton/ledger/error/CommonErrors$$UnsupportedOperation$.html>`_`

The part `../../canton/.....` means for Sphinx to take file location where this is stated, and then traverse from there when creating the link.

Versus `/canton/.....` means for Sphinx to create the link _relative_ to Sphinx root.

I'm extrapolating (guessing) this from the [`literalinclude` directive documentation](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-literalinclude) which states

> The file name is usually relative to the current file’s path. However, if it is absolute (starting with /), it is relative to the top source directory.

From what I've seen building docs.daml.com documentation locally _**it should work**_. But I could not really test it locally because running it on localhost. – If this breaks I roll it back; and will look for different approach.

yves-da:
I'll give another try to test this locally.

garyverhaegen-da:
So that seems to be the [syntax for external links](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#external-links) in RST, meaning Sphinx does not touch those paths at all and includes them verbatim in the generated HTML.

Whereas
```
* **Scaladocs**: `UNSUPPORTED_OPERATION <../../canton/scaladoc/com/digitalasset/canton/ledger/error/CommonErrors$$UnsupportedOperation$.html>`_
```

defines a _relative path_, where the path will be dynamically constructed by the browser upon clicking,

```
* **Scaladocs**: `UNSUPPORTED_OPERATION </canton/scaladoc/com/digitalasset/canton/ledger/error/CommonErrors$$UnsupportedOperation$.html>`_
```

defines an _absolute path_ which will be independent from the page on which it appears.

This means that (in my understanding) the first version will work as expected on both the root and versioned sites (assuming the link is "correct" to start with, of course), whereas the second form will always link to the "root" version of the site even if the user is currently looking at a versioned site.

garyverhaegen-da:
Out of curiosity, what was the problem you were trying to solve by changing those links? What currently doesn't work as expected?

PR #341, opened 2023-06-27
Phol/canton driver integration [DRV-217]

PHOL-DA:
Update the docs.daml.com to use canton driver repo for example e01 and e03 (Besu and Fabric) for the upcoming 2.7.0 release.

[Related PR in canton](https://github.com/DACH-NY/canton/pull/13460).


docs/2.7.0/docs/canton/usermanual/domains/ethereum.rst
```diff
@@ -202,7 +202,7 @@ Canton supports mutual TLS between Canton and Ethereum client nodes and the demo
 configure this.
 Concretely, the TLS configuration for Canton expects a key store and the path to the Ethereum TLS certificates:
 
-.. literalinclude:: /canton/includes/mirrored/enterprise/app/src/pack/examples/e03-ethereum-sequencer/canton-conf/simple/tls.conf
+.. literalinclude:: /canton-drivers/examples/e03-ethereum-sequencer/canton-conf/simple/tls.conf
```
yves-da, line 205: Will there be also an update to the canton repository?

Specifically, will these example be deleted from the canton repository?

If you need help with that removal, I can help.

PHOL-DA: i already have a PR where i remove the examples (but the pipeline line fails because the docs cant be generated without the). That is why im here :) 

Thanks for offering

PHOL-DA: i will make sure to update it accordingly aswell then. it is not only the examples being deleted, it is much more (associated driver's aswel)

docs/2.7.0/docs/canton/usermanual/domains/ethereum.rst
yves-da, line 1: Because this RST file is now part of the Driver's Team documentation I recommend that you move it into the `/canton-drivers` directory.

I would do the same for any other RST file which is now part of the Driver's Team documentation.

PHOL-DA: i would wait with doing that, there is a high level decision awaiting to be made whether we combine drivers and canton for release or not. So from a client side (at least until a decision is made) drivers are part of canton.

We will do a document update later on when a final decision has been made on the release artifacts.

docs/2.7.0/docs/canton/usermanual/domains/ethereum.rst
yves-da: There is another "problem". The RST files with the canton directory in this repo are used by the canton repository to build a "canton documentation" (verification). This happens also on CircleCI for canton.

With the current changes, e.g. `literalinclude:: /canton-drivers/examples/e03-ethereum-sequencer/canton-conf/simple/tls.conf`, the canton build will fail.

To fix that, either
* the canton-drivers RST files need to move outside of the canton directory in this repository, or 
* I will have to modify the canton build to somehow ignore these changes.

docs/2.7.0/docs/canton/usermanual/domains/ethereum.rst
yves-da: That however, does not enforce or imply that the RST files need to be located with the `canton` directory in the docs.daml.com repository.

It's probably even helpful to put them into the `/canton-drivers/` directory rather than having them in `/canton/usermanual/domains` directory.

docs/2.7.0/versions.json
```diff
@@ -1,5 +1,6 @@
 {
   "daml": "2.7.0-snapshot.20230523.11810.0.06abd5c6",
   "canton": "2.7.0-snapshot.20230623.10628.0.v51275048",
-  "daml_finance": "1.2.14.99.20230615.1"
+  "daml_finance": "1.2.14.99.20230615.1",
+  "canton_drivers": "0.1.9"
```
yves-da, line 5: This downloads about 292 MB while the to be included files are just a few kilobytes.

I recommend to create a docs-only assembly tarball which only contains what the docs.daml.com documentation build needs to build.

docs/2.7.0/docs/canton/usermanual/domains/ethereum.rst
PHOL-DA: hmm okay, so it actually makes more sense to move the `ethereum.rst` and `fabric.rst` domain usermanual out into it's own folder for canton-drivers ?

I was not aware of this tight coupling between the files

docs/2.7.0/docs/canton/usermanual/domains/ethereum.rst
garyverhaegen-da: I'd defer that decision to the @digital-asset/daml-docs team.

My personal opinion is that we should be careful not to inflict our internal team structure on our users, and file hierarchy should as much as reasonable reflect the structure of the toc.

docs/2.7.0/docs/canton/usermanual/domains/ethereum.rst
garyverhaegen-da: The canton build should not be putting undue constraints on the docs repo.

docs/2.7.0/versions.json
PHOL-DA: i have updated the tar ball we generate, once we do the next release it should only contain relevant files

yves-da:
@PHOL-DA, @garyverhaegen-da has some fair points, so let's leave the repository / content structuring to @digital-asset/daml-docs team.

yves-da (approved):
I merged a PR to the canton repository to cope with this change.

Changes in this PR look good to me.

PR #362, opened 2023-07-11
bump daml

garyverhaegen-da:


garyverhaegen-da:
This is mostly to get the updated copyright notice, but it's also just been a while since we bumped it.

loiskilberg-da (approved):

PR #363, opened 2023-07-12
Remove todo for issue 156

yves-da:
After fixing the canton todo checker, this todo will be reported and blocks PR merges on the canton repository. Because the related issue has been closed for sometime this todo is simply removed.

Relates to [canton issue 13862](https://github.com/DACH-NY/canton/pull/13862)

### DIFF
--- a/docs/2.7.0/docs/canton/architecture/requirements/requirements.rst
+++ b/docs/2.7.0/docs/canton/architecture/requirements/requirements.rst
@@ -127,9 +127,6 @@ and design limitations :ref:`later in the section
   can resend change requests in case of a restart without adding the
   changes to the ledger twice.
 
-  .. todo::
-      #. `Change request deduplication / disaster recovery <https://github.com/DACH-NY/canton/issues/156>`_
-
   .. _bounded-decision-time-hlreq:
 
 * **Bounded decision time**. I want to be able to learn within some


yves-da:
CC @garyverhaegen-da, `.. todo::` markers in the canton RST files get checked against open/closed canton issues as part of the canton build.

andreaslochbihler-da (approved):

garyverhaegen-da:
> CC @garyverhaegen-da, `.. todo::` markers in the canton RST files get checked against open/closed canton issues as part of the canton build.

Is that an unfortunate remnant of the time before the docs move, or is that something we want to keep moving forward? If the latter, do you have any guidelines for the docs for what they need to do around that?

yves-da:
> 



> Is that an unfortunate remnant of the time before the docs move, or is that something we want to keep moving forward? If the latter, do you have any guidelines for the docs for what they need to do around that?

It's something we want to keep. In particular, @andreaslochbihler-da requested that this facility is working as before as it helps him writing up documentation. Probably also other canton members will use it as they have used it heavily in the past.

As a guideline for the docs team, I think it's just "Leave `.. todo::` blocks in the canton RST files as is". – I'll write to the product-docs slack channel to let others know as well.

PR #375, opened 2023-07-18
Move Reference App section

matteolimberto-da:
I moved the reference to the [Daml Finance Reference App](https://github.com/digital-asset/daml-finance-app) to the main `intro` section.

It was previously sitting in the intro to the getting started tutorials.

I feel that the previous placement positioned the `Reference App` as a graphical-based alternative to the getting started scripts. However, it should not be the case as the Reference App is not documented to a sufficient extent.

Happy to know what you think


johan-da (approved):
LGTM

brian-weir (approved):

PR #377, opened 2023-07-19
Update docker example for new images

soren-da:
The new docker images don't have the simple-ping.canton anymore and the simple-topology.conf has been moved.

### DIFF
--- a/docs/2.7.0/docs/canton/usermanual/docker.rst
+++ b/docs/2.7.0/docs/canton/usermanual/docker.rst
@@ -33,17 +33,16 @@ Starting Canton
 
 The canton executable is the default image entry point so all examples using ``bin/canton`` can simply substitute that with ``docker run digitalasset/canton``.
 
-For example, to run our example simple topology:
+For example, to run with our simple topology configuration in interactive console mode:
 
 .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


garyverhaegen-da:
I don't feel I am in a position to review this, as I'm lacking a lot of context. My understanding of the company position is still that we should not provide Docker images for open-source versions of our components, as Docker images are an enterprise feature.

Ignoring that, I'm not aware of where the canton open-source Docker image is defined and how/why it has changed.

Tangentially related, I would point out that the commands a couple lines above of this change should be changed to

```
docker login digitalasset-docker.jfrog.io
docker pull digitalasset-docker.jfrog.io/canton-enterprise
```

but I'm unsure what else needs to change accordingly as I'm wholly unfamiliar with the documentation of the Canton Docker images.

garyverhaegen-da (approved):
With the caveat of my earlier comment, I can confirm that:

1. `docker run --rm -it digitalasset/canton-open-source:latest --config simple-topology.conf` works.
2. `docker run --rm -it digitalasset/canton-open-source:latest --config examples/01-simple-topology/simple-topology.conf --bootstrap examples/01-simple-topology/simple-ping.canton` does not work.

So let's merge this and resolve the other points later on.

PR #403, opened 2023-08-08
Remove tps

garyverhaegen-da:
Advice for reviewers: this PR is separated into three commits. The first one is just tweaking `.gitignore` files and can be mostly ignored. The second one just copies the relevant Canton files to this repo, so there's a lot of added text. This is a straight up copy from the Canton tarball, with no change. Finally, the last commit, which should be the only one with an impact on the live website, removes the offending sections from the four "imported" files.

### DIFF
--- a/docs/2.3.15/.gitignore
+++ b/docs/2.3.15/.gitignore
@@ -1,1 +1,2 @@
 .envrc.private
+workdir/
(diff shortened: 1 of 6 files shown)


loiskilberg-da (approved):
Looks good. Thank you so very much!!

PR #404, opened 2023-08-08
Fill placeholders for Generic instrument and Token instrument

matteolimberto-da:


docs/2.7.0/docs/daml-finance/instruments/generic.rst
```diff
@@ -1,5 +1,162 @@
 .. Copyright (c) 2023 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
 .. SPDX-License-Identifier: Apache-2.0
 
-Generic Instrument
-##################
\ No newline at end of file
+How to use the Generic Instrument packages
```
GeorgSchneider, line 4: Here and I think in other places we say "Instrument packages" (plural). Are there actually multiple ones, or should this be singular?

matteolimberto-da: Each instrument is "distributed" across two packages (the interface and the implementation), hence I've been using plural.

I was actually planning of adding a table for each instrument enumerating the two packages, wdyt?

GeorgSchneider: Nah, I think it's good. thanks

GeorgSchneider (approved):

PR #405, opened 2023-08-08
Documented to start Canton via java directly

soren-da:
Instead of always relying on `bin/canton` to start Canton, we now also document to start Canton via `java` in the getting started guide.

### DIFF
--- a/docs/2.7.0/docs/canton/tutorials/getting_started.rst
+++ b/docs/2.7.0/docs/canton/tutorials/getting_started.rst
@@ -46,7 +46,6 @@ The extracted archive has the following structure:
   ├── daml
   ├── dars
   ├── demo
-  ├── deployment
   ├── drivers (enterprise)
   ├── examples
   ├── lib
@@ -56,7 +55,6 @@ The extracted archive has the following structure:
 - ``daml``: contains the source code for some sample smart contracts
 - ``dars``: contains the compiled and packaged code of the above contracts
 - ``demo``: contains everything needed to run the interactive Canton demo
-- ``deployment``: contains a few example deployments to cloud or docker
 - ``examples``: contains sample configuration and script files for the Canton console
 - ``lib``: contains the Java executables (JARs) needed to run Canton
 
@@ -81,7 +79,11 @@ Then, run
 
    bin/canton --help
 
-to see the command line options that Canton supports.
+to see the command line options that Canton supports. Alternatively to
+`bin/canton` you can also start Canton directly with `java -jar
+lib/canton-*.jar` assuming all other jar dependencies are in the `lib` folder
+too.
+
 Next, run
 
 .. code-block:: bash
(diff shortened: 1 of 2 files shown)


marciw-da:
carrielaben-da is on PTO so added daml-docs for review instead.

garyverhaegen-da:
Could we specifically document the need for the BouncyCastle jar? The current documentation feels misleading, and I don't think it should rely on Rémy's hack. I'd much rather we document how to properly start things with `-cp`.

docs/2.7.0/docs/canton/tutorials/getting_started.rst
```diff
@@ -81,7 +79,11 @@ Then, run
 
    bin/canton --help
 
-to see the command line options that Canton supports.
+to see the command line options that Canton supports. Alternatively to
+`bin/canton` you can also start Canton directly with `java -jar
```
loiskilberg-da, line 83:
```suggestion
``bin/canton``, you can also start Canton directly with ``java -jar
```

docs/2.7.0/docs/canton/tutorials/getting_started.rst
```diff
@@ -81,7 +79,11 @@ Then, run
...
+lib/canton-*.jar` assuming all other jar dependencies are in the `lib` folder
```
loiskilberg-da, line 84:
```suggestion
lib/canton-*.jar``, assuming all other jar dependencies are in the ``lib`` folder,
```

docs/2.8.0/docs/canton/tutorials/getting_started.rst
```diff
@@ -81,7 +79,11 @@ Then, run
 
    bin/canton --help
 
-to see the command line options that Canton supports.
+to see the command line options that Canton supports. Alternatively to
+`bin/canton` you can also start Canton directly with `java -jar
```
loiskilberg-da, line 83:
```suggestion
``bin/canton``, you can also start Canton directly with ``java -jar
```

docs/2.8.0/docs/canton/tutorials/getting_started.rst
```diff
@@ -81,7 +79,11 @@ Then, run
...
+lib/canton-*.jar` assuming all other jar dependencies are in the `lib` folder
```
loiskilberg-da, line 84:
```suggestion
lib/canton-*.jar``, assuming all other jar dependencies are in the ``lib`` folder,
```

loiskilberg-da (approved):
Looks good! Just added some small formatting suggestions.

soren-da:
> Could we specifically document the need for the BouncyCastle jar? The current documentation feels misleading, and I don't think it should rely on Rémy's hack. I'd much rather we document how to properly start things with `-cp`.

Remy's approach is not a hack, but [standard way](https://docs.oracle.com/javase/tutorial/deployment/jar/downman.html) to specify the JAR's dependency in the manifest, and it works automatically with all fat jar excluded dependencies. Listing the BC jar explicitly and start canton via `java -cp` has a few issues that can break the docs in the future: 1) if the BC jar name changes e.g. due to a version upgrade, 2) if the Canton main class name changes, because we have to specify it when starting via `java -cp`, 3) if we need to add further fat jar excluded jars.

daravep (approved):
LGTM

PR #416, opened 2023-08-14
fix snapshot ignore

garyverhaegen-da:
@yves-da The snapshot ignore list is based on the overall snapshot version, which is the daml version the canton snapshot was built against, not the canton version itself.


yves-da (approved):
Oh! I see. Thank you for the explanation.

PR #420, opened 2023-08-14
make a change to trigger update

garyverhaegen-da:
It turns out the update script is a little bit broken: it only checks if the commit of the docs has changed, not the version. I'll fix it later, but for now let's get 2.7.0 out.

(What happened here is that the last [commit](https://github.com/digital-asset/docs.daml.com/pull/415) to touch 2.7.0 _also_ touched 2.6.5 so the commit was correct and it did not make any update.)


carrielaben-da (approved):
LGTM

PR #430, opened 2023-08-16
Use indicative mode in phrasing, switch to pg instead of h2

mziolekda:
Addresses one of the problems highlighted in https://github.com/DACH-NY/canton/issues/13566


oliverse-da (approved):
Looks good. Thank you, Marcin

PR #450, opened 2023-09-01
explain reference pattern

matteolimberto-da:
Fixes https://github.com/digital-asset/daml-finance/issues/1028


johan-da (approved):
LGTM

docs/2.8.0/docs/daml-finance/reference/patterns.rst
```diff
@@ -51,6 +51,35 @@ could in principle stay untouched.
...
+ - contains the contract id of the target instrument template (e.g. a fixed rate bond)
+
+We can then fetch an interface "by key" by 
+ - fetching the `Reference` template by key (``fetchByKey``)
+ - reading and ``fetch``-ing the stored contract id
+
+.. image:: ../images/reference_pattern.png
+   :alt: Example using the reference pattern to fetch an interface "by key".
+
+Similarly, this pattern lets us also exercise a choice on an interface "by key".
```
markus-da, line 77:
```suggestion
Similarly, this pattern also lets us exercise a choice on an interface "by key".
```

docs/2.8.0/docs/daml-finance/reference/patterns.rst
```diff
@@ -51,6 +51,35 @@ could in principle stay untouched.
...
+:ref:`InstrumentKey <constr-daml-finance-interface-types-common-types-instrumentkey-32970>` to
+identify instruments across a number of implementing templates.
+
+To do that, we define a `Reference` template that
+ - is keyed by the `InstrumentKey`
+ - contains the contract id of the target instrument template (e.g. a fixed rate bond)
+
+We can then fetch an interface "by key" by 
+ - fetching the `Reference` template by key (``fetchByKey``)
+ - reading and ``fetch``-ing the stored contract id
```
markus-da, line 72: Do we maybe want to explain where the interface comes in? We say `We can then fetch an interface "by key" by` but then only mention `Reference` and `contract id`.

markus-da (approved):
Left a couple of minor comments.

matteolimberto-da:
> Left a couple of minor comments.

Thanks Markus, will address them with a separate PR

docs/2.8.0/docs/daml-finance/reference/patterns.rst
matteolimberto-da: It is shown in the picture (the Reference template contains the `cid : ContractId Instrument`) but you are right that we don't specifically mention the interface in the wording

docs/2.8.0/docs/daml-finance/reference/patterns.rst
```diff
@@ -51,6 +51,35 @@ could in principle stay untouched.
...
 :doc:`Holdings tutorial <../tutorials/getting-started/holdings>`.
 
+.. _reference-pattern:
+
+Reference pattern
+-----------------
+
+The Reference pattern is used to leverage the functionalities of :ref:`Contract Keys <contractkeys>`
+when working with interfaces.
+
```
markus-da, line 61: Maybe also clarify "There is currently no built-in support for interface keys, so this is used as a workaround", in order to specify why we are using this pattern and what would happen if we would not use it.

PR #477, opened 2023-09-22
update from postgres 11 to postgres 14 in docker instructions

dasormeter:
Since Canton and PQS both support postgres 14, update to this newer version in the examples and docs


marciw-da (approved):
LGTM and thanks for the quick action on this!

PR #479, opened 2023-09-24
Domain language updates part 2

carrielaben-da:
Primarily in the Canton docs.


loiskilberg-da (approved):
Thanks for all the edits!!!! I made just a few suggestions. Looks good!

PR #489, opened 2023-10-02
simplify Python install

garyverhaegen-da:
Specifically, if Python is already set up, don't reinstall it from scratch every time one `cd`'s into the folder.


loiskilberg-da (approved):
thank you

PR #494, opened 2023-10-05
fix snapshots for assembly change

garyverhaegen-da:
Daml version is no longer tied to SDK version.


loiskilberg-da (approved):
thank you!

PR #505, opened 2023-10-19
document the backup order for the domain nodes

andreaslochbihler-da:


docs/2.8.0/docs/canton/usermanual/persistence.rst
```diff
@@ -563,6 +563,18 @@ The state of applications that interact with participant's ledger API must be
...
 reset.
 
+Among the different nodes of a domain (sequencer, mediators, domain topology manager),
+the following constraints apply:
+
+- Back up the mediators and the domain manager before the sequencer;
+  otherwise they may not be able to reconnect to the sequencer (``ForkHappened``).
+  The relative order of mediators, domain topology manager, and participants does not matter.
+
+- On a :ref:`permissioned domain <permissioned-domains>`, make sure that
```
rgugliel-da, line 573: How would someone guarantee that ?

rgugliel-da (approved):

docs/2.8.0/docs/canton/usermanual/persistence.rst
andreaslochbihler-da: In my view of the world, the operator of the domain topology manager is the same as the operator of the sequencer. That's the trust model in Daml 2. Moreover, only the operator should have access to the domain topology manger's admin api that's needed to add the participant. So this operator should be able to ensure this by themself. For example, by preventing access to the particular admin endpoint between the two backups.

rgugliel-da: I see that but we could argue that a backup should be done periodically automatically, which makes the above requirement difficult to enforce

carrielaben-da (approved):
LGTM

docs/2.8.0/docs/canton/usermanual/persistence.rst
andreaslochbihler-da: Well, if you use a cron job for the backup, then use the same cron job to block access to the admin api. If your team has too little work, we can try to look into how to avoid this constraint. Otherwise, I'd leave this as is until a customer complains loudly.

PR #518, opened 2023-11-01
remove generated theme files

garyverhaegen-da:
I'm not entirely sure how they got committed as they are included in the `.gitignore` file already.


marciw-da (approved):
yikes and thanks :)

PR #522, opened 2023-11-05
remove the union operation when determining the fetch actors

yilin-sai:
According to the discussion [here](https://discuss.daml.com/t/fetch-a-contract-as-a-non-stakeholder/6915/5?u=sai), I think DAML is not doing the union operation when determining the fetch actors. This PR is to correct the docs by removing the union operation.

It would be good if someone with the expertise and time can confirm that the union is not performed in DAML source code. The DAML template and script [here](https://discuss.daml.com/t/fetch-a-contract-as-a-non-stakeholder/6915?u=sai) can be considered as a blackbox test.


yilin-sai:
Just figured out that the union is useful when the controller is a divulgee of c2, so the blackbox test is missing a divulgence step. Sorry for the confusion, I'm closing this PR.

PR #540, opened 2023-11-20
Rename references according to files being moved in the Canton repo

stefanobaghino-da:


stefanobaghino-da:
@mziolekda Thanks, but this requires an approval from @digital-asset/daml-docs. 🙂

mziolekda:
I realize that. But at least I have checked the paths with respect of the canton source and can vouch for their soundness. 
If that helps in getting the final approval from the docs team

stefanobaghino-da:
I temporarily made this into a draft since we're evaluating the current impact of those path changes. If we confirm we want to move forward, I'll turn this back into a merge-able PR once we can confirm we want to move on with this. I would still appreciate a review so that, if confirmed, we can simply merge.

loiskilberg-da (approved):
Thanks!

stefanobaghino-da:
As mentioned above, unforeseen consequences forced us to roll back the change that made this PR necessary. Closing. Thanks for taking the time anyway, luckily it was trivial. 🙇🏻‍♂️

PR #543, opened 2023-11-20
Create SECURITY.md

nycnewman:
Add a SECURITY file to repo

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,19 @@
+# Security Policy
+
+## Supported Versions
+
+The latest released version of Daml is supported with security updates.
+You can find the latest version at:
+https://docs.daml.com/support/release-notes.html
+
+## Reporting a Vulnerability
+
+If you suspect that your issue has a security impact, do not use the issue tracker.
+In that case, please follow our responsible disclosure policy:
+https://www.digitalasset.com/responsible-disclosure
+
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -0,0 +1,19 @@
...
+
+The latest released version of Daml is supported with security updates.
+You can find the latest version at:
+https://docs.daml.com/support/release-notes.html
+
+## Reporting a Vulnerability
+
+If you suspect that your issue has a security impact, do not use the issue tracker.
+In that case, please follow our responsible disclosure policy:
+https://www.digitalasset.com/responsible-disclosure
```
marciw-da, line 13: This confused me slightly because "the issue tracker" hasn't been mentioned before now. But I suppose in most cases people will get to this policy via the issue tracker, so in that sense it's been mentioned. :) 

Up to you. If you do want to make it a bit clearer, you could swap this paragraph with the next paragraph and change the heading to "Reporting an Issue or Vulnerability"

marciw-da (commented):
Thanks for adding this file! Some small suggestions.

We will also add a link or two in the Daml docs

SECURITY.md
marciw-da: I understand that; I'm just saying it hasn't been mentioned earlier on this page so it kind of came out of nowhere. Not a big deal.

marciw-da (approved):

PR #387, opened 2023-07-27
add 2.8.0 folder for docs

loiskilberg-da:
Copied 2.7.0 docs folder to become a new 2.8.0 docs folder. Did not touch versions.json, dropdown_versions, or root files.

PR #544, opened 2023-11-21
Add 3.0.0 directory for docs

yves-da:
Required for https://github.com/DACH-NY/canton/issues/15204

As before, https://github.com/digital-asset/docs.daml.com/pull/387, this copies the current `2.8.0` docs directory to a new `3.0.0` directory.


garyverhaegen-da:
Note that this means, until 2.8.0 gets published at least, most docs changes now need to be replicated in three places.

yves-da:
> Note that this means, until 2.8.0 gets published at least, most docs changes now need to be replicated in three places.

Indeed. – As we're going to remove *a lot* of Canton code and change things around it's likely it'll affect the snippets/source code includes so having 3.0.0 now is the better option.

loiskilberg-da (approved):

PR #546, opened 2023-11-21
Update intro.rst

frankPreiwussDA:
I deleted this paragraph about future development. We will communicate roadmap topics via other channels, then we avoid the risk that we need to change it later again.


loiskilberg-da (approved):
Thank you!

yves-da:
@frankPreiwussDA – There's now also an unpublished `3.0.0` version. It'd be helpful if you apply these changes in both versions, `2.8.0` and `3.0.0`. Thank you.

johan-da:
@frankPreiwussDA applied the change to the 3.0.0 fork as well.

PR #563, opened 2023-12-04
Updated Kubernetes target deployment diagram to exclude 'passive' nodes [PLEN-548]

miklos-da:
Summary of changes:
* Updated Kubernetes target deployment diagram to exclude 'passive' nodes, as it is not relevant in a Kubernetes context.
* Referenced the section on HA on Kubernetes from the overview.


loiskilberg-da (approved):
thank you!

PR #566, opened 2023-12-05
Restructure Assistant section

marciw-da:
[do not review until Wednesday!!] ☃️ 

Restructure the Assistant section to accommodate some new documentation on multi-package builds (coming soon in a separate PR from @samuel-williams-da). The new content would make the page quite long and unwieldy if we were to just add it to the existing page.

* Break up the page and create new sub-pages with the existing content
* Add the new pages to the ToC
* Make a few edits to ToC entries and headings (probably more coming in a later PR after some team consultation)

### DIFF
--- a/docs/2.8.0/_toc.yml
+++ b/docs/2.8.0/_toc.yml
@@ -249,6 +249,16 @@ subtrees:
       entries:
       - file: tools/assistant
         title: "Daml Assistant"
+        subtrees:
+        - hidden: False
+          titlesonly: True
+          entries:
+          - file: tools/assistant-install
+            title: "Manage Versions"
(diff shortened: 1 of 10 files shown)


marciw-da:
I just looked at the description of this PR and realized it's pretty unclear what I'm trying to do! 😒 

These changes are purely structural, no new content and nothing to edit. My only goal was to split the page into multiple pages to prepare the way for new content.

So for now it's just "does this structure look right"? I haven't edited any of it for style or clarity yet but am hoping to do that as part of incorporating the new content. (If you have already started a review, I will of course welcome an editing pass now -- just clarifying that I'm not expecting that.) 🙇

PR #585, opened 2023-12-14
Code fork, 3.x: port `RehydrationIntegrationTest`-related docs to X

fabiotudone-da:
Companion PR of https://github.com/DACH-NY/canton/pull/15823.

@digital-asset/daml-docs This section is essentially "JFYI" and anyway will evolve before 3.x because the UX of these tools has room for improvement. In any case, it is simpler now and I tried to reflect it in the docs.


loiskilberg-da (approved):

loiskilberg-da:
Note that very soon, the 3.0.0 docs folder will be copied to a new 2.9.0 folder. If this PR doesn't get merged first, you may need to copy it manually to 2.9.0.

PR #592, opened 2023-12-19
Rewrote installation guide to use new reference config (#583)

daravep:
Ported config / docs changes of #583  to 2.8 

Before, the installation guide was based on the advanced config example, which was guiding people through the process by starting with the simplest example and then calling out options.

Now, we start with a full example that comes with TLS, JWT, Storage etc, where a client can then decide to incrementally remove configurations instead.

The docs changes are related to https://github.com/DACH-NY/canton/pull/15882


marciw-da (approved):
Thanks for working through all the merge challenges and for backporting this.

PR #595, opened 2023-12-21
Added inconsistent snapshot release to list of broken snapshots

daravep:
Publishing the snapshot 2.9.0-snapshot.20231220.0 docs failed because the docs in the repository can not be built with canton 2023-12-15, but require at least 2023-12-19.


dasormeter (approved):

marciw-da (approved):
Thanks for the fix.

PR #600, opened 2024-01-04
remove dead link

garyverhaegen-da:
Nobody seems to know what this is meant to link to. Plus, as far as I can tell this is the only page on which we talk about metrics, so we don't really need a link.


garyverhaegen-da:
Note that this is already broken in [the 2.6.5 docs](https://docs.daml.com/2.6.5/canton/usermanual/monitoring.html#prometheus-only-metrics) though the page it's trying to link to [does exist in 2.6.5](https://docs.daml.com/2.6.5/ops/common-metrics.html), but apparently got merged at some point in the 2.7 line.

garyverhaegen-da:
cc @benm-da @michael-godfrey-da who have both reported this.

carrielaben-da (approved):
LGTM

PR #604, opened 2024-01-08
Updated support policy

daravep:
Before this change, we didn't mention the support policy as per the new agreement. Now, we do. https://docs.google.com/document/d/14IS8ECgZWV9BEXBz1BU1gEwWaYsesTmNSbtoGPXAWGE/edit#heading=h.ydftn3ns7yuk


bame-da (approved):
ty

marciw-da (approved):
Approving as is because we have a separate spruce-up planned for this page.

cc @shaul-da who wrote the previous version fairly recently

PR #614, opened 2024-01-17
pin Sphinx deps

garyverhaegen-da:
Sphinx 4.3.1 depends on "most recent applehelp", which since Aug 2023 requires Sphinx 5, so our current install is broken. (Same applies to a handful of other Sphinx deps.)

It took some time to surface as the old install was cached by CircleCI (and, likely, it's not affected anyone's local builds either for the same reason).


daravep (approved):
LGTM

carrielaben-da (approved):

PR #622, opened 2024-01-23
Various small tweaks

garyverhaegen-da:
Inspired by a conversation on Slack.

### DIFF
--- a/docs/2.6.5/docs/daml/intro/1_Token.rst
+++ b/docs/2.6.5/docs/daml/intro/1_Token.rst
@@ -31,7 +31,7 @@ Individual contracts are *immutable* in the sense that an active contract can no
 
 Daml specifies what transactions are legal on a Daml Ledger. The rules the Daml code specifies are collectively called a *Daml model* or *contract model*.
 
-Daml Files and Modules
+Daml Modules and Files
 ----------------------
 
 Each ``.daml`` file defines a *Daml Module* at the top:
--- a/docs/2.6.5/docs/index.rst
+++ b/docs/2.6.5/docs/index.rst
@@ -1,0 +1,2 @@
+Daml Documentation
+------------------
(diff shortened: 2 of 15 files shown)


docs/2.6.5/docs/daml/intro/1_Token.rst
```diff
@@ -31,7 +31,7 @@ Individual contracts are *immutable* in the sense that an active contract can no
 
 Daml specifies what transactions are legal on a Daml Ledger. The rules the Daml code specifies are collectively called a *Daml model* or *contract model*.
 
-Daml Files and Modules
+Daml Modules and Files
```
garyverhaegen-da, line 34: This page probably needs a bit of an editing pass as I'm now realizing this is not the only discrepancy between the list of titles and the actual titles, but this is the one I had noticed.

docs/2.6.5/docs/index.rst
```diff
@@ -0,0 +1,2 @@
+Daml Documentation
+------------------
```
garyverhaegen-da, line 2: This fixes the title of the "root" page to be `Daml Documentation` instead of the current `Daml Files` (which comes from the daml repo bundle's fake index page).

theme/docs/conf.py
```diff
@@ -3,7 +3,7 @@
 
 # -*- coding: utf-8 -*-
 #
-# Daml SDK documentation build configuration file, created by
+# Daml documentation build configuration file, created by
```
garyverhaegen-da, line 6: These are default values currently overridden by the `config.py` files coming from the daml repo bundle. Still, I thought updating them made sense.

docs/2.6.5/docs/daml/intro/1_Token.rst
marciw-da: Thanks for the heads-up -- will do that in a separate PR

marciw-da (approved):
Thank you for the quick fixes!

PR #633, opened 2024-02-05
fix small spelling issue

garyverhaegen-da:


garyverhaegen-da:
> "Set up" should be two words as a verb, if you want to fix that in the meantime (but I'd call it optional). https://developers.google.com/style/word-list#setup

Oh. I hadn't read beyond the three words `a service agreements`; that somehow caught my eyes when the text scrolled while grepping for something else.

PR #634, opened 2024-02-05
simplify nix setup

garyverhaegen-da:
This started with 2.4.3 failing to build in #633. Digging into the error, it turns out 2.4.3 is built with a very old version of Nix, which comes with a trusted root CA store that's too old to recognize GitHub's certificate, which in turn fails the build of the theme.

The most direct solution would be to update the Nix definition, but unfortunately 2.4.3 gets its Sphinx version from Nix, so that would also mean bumping Sphinx itself, which looked like a lot of work. Or changing the Nix setup of 2.4.3 to get Sphix from a different channel than the SSL certs. That seemed messy.

Or, maybe the right way to solve this would be to change the Nix setup of 2.4.3 to match later ones, i.e. use Nix to get Python only, then use Python to get the version of Sphinx we want.

But that got me thinking: why do we have different versions of Nix for different versions of the docs? I don't remember that being an explicit decision; I think it was just an artefact of switching from branches to folders, way back when.

So this PR resolves the 2.4.3 build failure in #633 by unifying all the Nix configurations in this repo into a single one, which also simplifies the CircleCI config a bit.

While I was "in there" I also changed the CircleCI configuration to update the version of Nix itself and the version of the base Docker image we use.


gerolf-da (approved):
Thanks for the cleanup 👍

marciw-da (approved):

PR #556, opened 2023-11-29
Explain cron format for pruning schedules

stefanobaghino-da:

PR #672, opened 2024-03-26
Add Schedule format

marciw-da:
Backporting new Schedule format section from 3.0.0 -- see also #556

### DIFF
--- a/docs/2.9.0/docs/canton/usermanual/pruning.rst
+++ b/docs/2.9.0/docs/canton/usermanual/pruning.rst
@@ -3,6 +3,10 @@
 ..
    Proprietary code. All rights reserved.
 
+.. |br| raw:: html
+
+   <br />
+
 .. _ledger-pruning:
 
 Ledger Pruning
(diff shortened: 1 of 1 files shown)


garyverhaegen-da (approved):

docs/2.9.0/docs/canton/usermanual/pruning.rst
```diff
@@ -40,9 +44,81 @@ Refer to the cron specification to customize the pruning schedule. Here are a fe
...
+   * - day of the week
+     - number from ``1`` to ``7`` (``1`` = Sunday) 
+       |br|
+       **or** 
+       |br|
+       the first three letters of the day name (``SUN``, ``MON``, ``TUE``, ...)
+   * - year (optional)
+     - number from ``1900`` to ``2099``
+
+*\*Ranges specified with "to" are inclusive of both endpoints.*
```
garyverhaegen-da, line 84: This sentence threw me off - it took me a while to figure out that it refers to the ranges specified in the table above, and not to ranges the user would define.

How bad would it be layout-wise to just say "included" in all relevant entries in the table directly? (And then remove this sentence.) Alternatively,

```suggestion
*\*Ranges specified with "to" in the table above are inclusive of both endpoints.*
```

marciw-da: Hmm. Were you looking at just the .rst or also the rendered view? There's an asterisk that connects the column heading to the note (you can also see it here https://docs.daml.com/3.0.0/canton/usermanual/pruning.html#schedule-format)

Maybe I'll go with "In the table, ranges specified..." 

table

marciw-da: I can also look at putting the note right there under the column heading, though I think I tried that and it was messy. But I'll check.

garyverhaegen-da: Ah, sorry, no, I was just reading through the rst. It is indeed a lot clearer with the layout.

PR #675, opened 2024-03-28
Restart 3

garyverhaegen-da:
Reviewing most file changes here is inefficient; here's a high-level description:

1. Remove `docs/3.0.0`. This folder has been too abused over time to be a good starting point for anything.
2. Copy `docs/2.9.0` to `docs/3.1`.
3. Update `docs/3.1/README.md` to be a bit closer to the current situation.

Note that `docs/3.1/versions.json` has _not_ been updated at this point as that may induce many more changes that would make this PR even harder to review.


marciw-da (approved):
👏 🙏

rgugliel-da:
It is expected that the directory is named 3.1 and not 3.1.0 ?

PR #677, opened 2024-03-28
[pqs] spruce up code samples

garyverhaegen-da:
This is mostly changing the highlight settings on code blocks - plus a couple rewordings as it looks like our CSS is broken for `console` output (it's dark grey on a black background, not ideal if we want humans to read it).

cc @hrischuk-da as the main author of this page I believe.


carrielaben-da (approved):
LGTM

PR #689, opened 2024-04-18
downloads page styling

garyverhaegen-da:
No actual content yet, but this PR goes from looking like this:

Screenshot 2024-04-18 at 17 51 45
Screenshot 2024-04-18 at 17 51 59

to looking like this:
Screenshot 2024-04-18 at 17 51 48
Screenshot 2024-04-18 at 17 51 53


carrielaben-da (approved):

dylant-da (approved):
I've built and tried it out on my system, works great! Thank you for fighting your way through HTML/JS for this.

Three small suggestions, entirely optional:
- Adding a bit of padding to the bottom of the page, so that the bottom of the table doesn't mate exactly with the bottom of the browser?
- A drop shadow on the dropdown such as `box-shadow: rgba(0, 0, 0, 0.24) 0px 3px 8px;`
- Make it possible to click outside the dropdown to dismiss it, which is the normally behaviour for dropdowns

PR #695, opened 2024-04-29
Switch to using canton originated error categories and codes for 3.1

simonmaxen-da:


cathyjung-da (approved):

yves-da (commented):
Few thoughts on this change ✋ 

First, I'm not clear on state of the 3.1 version of the docs. From the canton side, my last info was that we do not work on the version 3 docs until we sort things out on how to do it. So this clean up may or may not make sense in that context.

Independent of that, I would attempt to remove the unwanted duplication also for the 2.* docs versions; at least include the 2.9 docs version as this is going to be the basis for the LTS version.

Regarding `docs/3.1/docs/canton/reference/canton-error-categories-inventory.rst.inc`:
The current mechanism is that generated content (like Canton error codes) is included in a tar-ball on which then the docs build depends on instead of generating the content and then checking it in the docs repository. – The motivation for that approach is to ensure that the checked-in, generated content does not mismatch with the software version it is supposed to document.
The experience with having a need to update checked-in generated content (like the Canton snippets) is bad. It gets easily forgotten, and then the documentation is out-of-sync.

docs/3.1/docs/canton/reference/error_codes.rst
```diff
@@ -320,20 +320,13 @@ in a sensible way to automatically deal with errors and decide whether to retry
...
-
-Error Codes Inventory - Daml
-****************************
-
-..
-   This file is generated:
-.. include:: error-codes-inventory.rst.inc
+   This file is manually generated using ErrorCategoryInventoryDocsGenApp the checked in
+   This was done as the daml variant is not maintained and due for removal
+.. include:: ./canton-error-categories-inventory.rst.inc
```
yves-da, line 325: I'm not clear on the comment.

Does `ErrorCategoryInventoryDocsGenApp` still exist? And do we need to continue to use it to generate the content for `canton-error-categories-inventory.rst.inc`?

If so, see my general comment about checking in generated content.

simonmaxen-da: Yes. My original proposal was to add this to the canton distribution and use this file with docs. I was told to hold off doing that for a few weeks as it was/is apparently unclear if the docs.daml.com repo will be merged back into the canton repo.

Once there is clarity regarding the above it should be done and I've raised an issue (https://github.com/DACH-NY/canton/issues/18872) to track this.

On the 2.9 branch this cannot be so easily done as these error codes are referenced by the ledger-api-test-tool (that in 3.x has been moved to the canton repo).

PR #697, opened 2024-05-01
Page title and TOC alignment

cathyjung-da:
The page titles and table of contents are aligned in the first half of the document, while the other half is a work in progress.


docs/2.9.0/docs/daml/intro/13_Interfaces.rst
```diff
@@ -1,8 +1,8 @@
 .. Copyright (c) 2023 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
 .. SPDX-License-Identifier: Apache-2.0
 
-Daml Interfaces
-===============
+Interfaces
+==========
```
carrielaben-da, line 5: "Daml Interfaces" is more information

docs/2.9.0/docs/getting-started/index.rst
```diff
@@ -3,8 +3,8 @@
 
 .. _new-quickstart:
 
-Getting Started with Daml
-#########################
+Getting Started
+###############
```
carrielaben-da, line 7: with Daml should remain here.

docs/2.9.0/docs/getting-started/installation.rst
```diff
@@ -1,8 +1,8 @@
 .. Copyright (c) 2023 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
 .. SPDX-License-Identifier: Apache-2.0
 
-Installing the SDK
-##################
+Installation
+############
```
carrielaben-da, line 5:
```suggestion
Install the SDK
##############
```

docs/2.9.0/docs/getting-started/testing.rst
```diff
@@ -1,8 +1,8 @@
 .. Copyright (c) 2023 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
 .. SPDX-License-Identifier: Apache-2.0
 
-Testing Your Web App
-********************
+Testing Your App
+****************
```
carrielaben-da, line 5:
```suggestion
Test Your App
*************
```

docs/2.9.0/docs/json-api/production-setup.rst
```diff
@@ -1,7 +1,7 @@
 .. Copyright (c) 2023 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
 .. SPDX-License-Identifier: Apache-2.0
 
-Production Setup
-################
+Overview
+########
```
carrielaben-da, line 5: Should be Production Setup

carrielaben-da (changes requested):
Overall, where there's a mis-match, we want to move both the titles and the TOC in the direction of providing more information. Mostly this is good but the instances I flagged should provide an example of what we're aiming for.

docs/2.9.0/docs/daml/intro/13_Interfaces.rst
cathyjung-da: Since an interface is something to be defined like a template, function, or a custom record type, wouldn't it be better to call it "Interface" instead of "Daml Interface"? It's like calling a template "Daml Template" a function each time.
Screenshot 2024-05-02 at 10 20 14 AM

docs/2.9.0/docs/getting-started/index.rst
cathyjung-da: Do we have standards for when to add "Daml" to the title and when not to? Since this entire document is dedicated to Daml, I find it unclear when we should or shouldn't include "Daml" in the title without making it sound redundant.

carrielaben-da: That seems like a good solution.

carrielaben-da (approved):

PR #713, opened 2024-06-05
Update 2.8 docs to 2.8.7

simonmaxen-da:


carrielaben-da (approved):

docs/2.8.7/versions.json
```diff
@@ -0,0 +1,6 @@
+{
+  "daml": "2.8.6",
```
yves-da, line 2: This should be `2.8.7` for daml and canton, no?

yves-da (changes requested):
Please double-check the versions in the versions.json file. Is it as intended?

docs/2.8.7/versions.json
garyverhaegen-da: No, component versions here. We have a Daml Enterprise 2.8.7, but sdk, canton, and drivers are all still at 2.8.6.

yves-da (approved):
@simonmaxen-da I'll approve the PR tentatively, because I don't want to block you. Please have a look at my comment, though.

PR #718, opened 2024-06-12
toc-title-matching-3

cathyjung-da:
TOC and page titles are aligned based on the changes merged to the 2.9 doc.


carrielaben-da (approved):
LGTM. However I would not change the titles/TOC entries of the Daml Finance docs since we don't manage those.

PR #740, opened 2024-07-04
Update docs section <Reduce the Number of Views>

JoaoSa-da:


docs/2.9.0/docs/canton/usermanual/performance.rst
```diff
@@ -506,15 +506,27 @@ If you submit a ledger API command (which can have multiple Daml commands), then
 
 Just putting all Daml commands into a single batch command does not help, as you might still create lots of views.
 
+Starting from protocol version 6, Canton creates a view for every action node in the transaction tree if the participants
+that host their informees are not a subset of their parents' informee participants. Therefore, in order to
```
andreaslochbihler-da, line 510: It's important that we look at the parent view, not at the parent action.
```suggestion
Starting from protocol version 6, Canton creates a view for every action node in the transaction tree if the participants
that host their informees are not a subset of their parent view's informee participants. Therefore, in order to
```

docs/2.9.0/docs/canton/usermanual/performance.rst
```diff
@@ -506,15 +506,27 @@ If you submit a ledger API command (which can have multiple Daml commands), then
...
+minimize the number of views, you should try to minimize either the number of changes of informees or make sure
```
andreaslochbihler-da, line 511: Informee changes are actually fine. What you want to avoid is "adding" informees in child nodes.
```suggestion
reduce the number of views, you should try to reduce either the number of times where the set of informees grows or make sure
```

docs/2.9.0/docs/canton/usermanual/performance.rst
```diff
@@ -506,15 +506,27 @@ If you submit a ledger API command (which can have multiple Daml commands), then
...
+most of your actions share the same pool of participants that must be informed. Another alternative is to
```
andreaslochbihler-da, line 512:
```suggestion
most of your actions share the same pool of participants that must be informed. Alternatively,
```

docs/2.9.0/docs/canton/usermanual/performance.rst
```diff
@@ -506,15 +506,27 @@ If you submit a ledger API command (which can have multiple Daml commands), then
...
+make your higher level action nodes include as much participants as possible so that subsequent action nodes in the tree
```
andreaslochbihler-da, line 513:
```suggestion
add informees to the action nodes close to the root (e.g., by including choice observers) so that their children nodes
```

docs/2.9.0/docs/canton/usermanual/performance.rst
```diff
@@ -506,15 +506,27 @@ If you submit a ledger API command (which can have multiple Daml commands), then
...
+are aggregated in the initial view, since these children nodes will only target a subset of those participants.
+Informees are roughly the signatories, observers of the contract and of the choice, and the controllers of the choice.
```
andreaslochbihler-da, line 515: They are defined in the ledger model's privacy section. I don't think it's useful to give a rough definition if you really want to optimize something.

docs/2.9.0/docs/canton/usermanual/performance.rst
```diff
@@ -506,15 +506,27 @@ If you submit a ledger API command (which can have multiple Daml commands), then
...
+
 In the protocol up to version 5, Canton creates a view for every action node in the transaction tree if the informees change
```
andreaslochbihler-da, line 517:
```suggestion
In the protocol up to version 5, Canton creates a view for every action node in the transaction tree if the informees or confirming parties change 
```

docs/2.9.0/docs/canton/usermanual/performance.rst
```diff
@@ -506,15 +506,27 @@ If you submit a ledger API command (which can have multiple Daml commands), then
...
+most of your actions share the same pool of participants that must be informed. Another alternative is to
+make your higher level action nodes include as much participants as possible so that subsequent action nodes in the tree
+are aggregated in the initial view, since these children nodes will only target a subset of those participants.
+Informees are roughly the signatories, observers of the contract and of the choice, and the controllers of the choice.
+
 In the protocol up to version 5, Canton creates a view for every action node in the transaction tree if the informees change 
-between the node and its parent. Therefore, in order to minimize the number of views, you should try to minimize the number of 
-changes of informees. Informees are roughly the signatories, observers of the contract and of the choice, and the controllers of the choice. 
+between the node and its parent. Therefore, in this case, in order to minimize the number of views, your only alternative
+is to try to minimize the number of changes of informees.
```
andreaslochbihler-da, line 519:
```suggestion
between the node and its parent. Therefore, in this case, to reduce the number of views, you must reduce the number of changes of informees and confirming parties.
```

docs/2.9.0/docs/canton/usermanual/performance.rst
```diff
@@ -506,15 +506,27 @@ If you submit a ledger API command (which can have multiple Daml commands), then
...
+that cover the exercise's informee participants, or a single stakeholder if your are using protocol version
+5 or older. You then send one exercise command on a generator contract with the list of contracts that in
+turn exercises the `Foo` choice per stakeholder group.
 
-If you write batch commands, group the operations on the ledger based on the informees and create a single choice and 
-batch operation per informee group. You can use choice observers to align informees if necessary. 
+If you write batch commands, group the operations on the ledger based on the informees and their hosting participants
+and create a single choice and batch operation per informee's participant group.
+For older protocol versions (v5<) that only take in consideration the informees and not their hosting participants, you can
+use choice observers to align informees if necessary.
```
andreaslochbihler-da, line 529: I think this test would be easier to follow if you introduce two sections here: One for PV <= 5 and one for PV 6 and greater. There may be a bit of repetition between the two sections, but readers should know what protocol version they are targeting and can then focus on that section. As is, they always have to evaluate these conditionals in their head in every sentence.

docs/2.9.0/docs/canton/usermanual/performance.rst
```diff
@@ -557,15 +571,19 @@ will have a huge impact on the number of views created.
...
+         -- is hosted by any of the contract's informee participants.
          observer obs
          controller owner 
          do
          forA_ batch (`exercise` Foo)
 
 
-As a rule, the number of views should depend on the number of stakeholder groups you have in your batch choice, 
-not the number of "batches" you process in parallel within one command.
+As a rule, the number of views should depend on the number of participant stakeholder groups you have in your batch choice
```
andreaslochbihler-da, line 582: You haven't defined what a participant stakeholder group is.

carrielaben-da (approved):
LGTM

docs/2.9.0/docs/canton/usermanual/performance.rst
```diff
@@ -506,74 +506,104 @@ If you submit a ledger API command (which can have multiple Daml commands), then
...
-In the protocol up to version 5, Canton creates a view for every action node in the transaction tree if the informees change 
-between the node and its parent. Therefore, in order to minimize the number of views, you should try to minimize the number of 
-changes of informees. Informees are roughly the signatories, observers of the contract and of the choice, and the controllers of the choice. 
+Strategy for Protocol Version 6
++++++++++++++++++++++++++++++++
 
-Instead of sending a batch with 20 exercise commands `Foo`, group your contracts by stakeholder and send one exercise 
-command on a generator contract with the list of contracts that in turn exercises the `Foo` choice per stakeholder group.
+Starting from protocol version 6, Canton creates a view for every action node in the transaction tree if the
+**participants that host their informees are not a subset of their parent view's informee participants**. Therefore, to
```
andreaslochbihler-da, line 513: Can you define the term "informee participant" first? AFAIK we have not defined this anywhere in our documentation. Since the term appears several times in this section, it might be worth to define it, say as, "the participants that must be informed about the view".

docs/2.9.0/docs/canton/usermanual/performance.rst
```diff
@@ -506,74 +506,104 @@ If you submit a ledger API command (which can have multiple Daml commands), then
...
+reduce the number of views, you should try to reduce either the number of times the set of informees grows or
```
andreaslochbihler-da, line 514:
```suggestion
reduce the number of views, you should try to reduce either the number of times the set of informee participants grows or
```

docs/2.9.0/docs/canton/usermanual/performance.rst
```diff
@@ -506,74 +506,104 @@ If you submit a ledger API command (which can have multiple Daml commands), then
...
 
-If you write batch commands, group the operations on the ledger based on the informees and create a single choice and 
-batch operation per informee group. You can use choice observers to align informees if necessary. 
+Instead of sending a batch with 20 exercise commands `Foo`, group your contracts by a set of
+stakeholders that cover the exercise's informee participants and send one exercise command on a generator contract
+with the list of contracts that in turn exercises the Foo choice per stakeholder group.
 
-For example, if the below `doUpdate` is only visible to the `owner` party, then whether the `efficient` flag is set or not 
+In general, if you write batch commands, group the operations on the ledger based on the informees' participants and create
+a single choice and batch operation per informee's participants group.
```
andreaslochbihler-da, line 524: You probably want to add a few sign posts to help the reader understand the flow here. The previous paragraph talks about reducing the number of views generated for children of a node (i.e., parent-child views). This paragraph now talks about batches of commands (i.e., sibling views). This is not clear from the text yet.

docs/2.9.0/docs/canton/usermanual/performance.rst
```diff
@@ -506,74 +506,104 @@ If you submit a ledger API command (which can have multiple Daml commands), then
...
+
+For example, if the `doUpdate` is only visible to the `owner` party, then whether the `efficient` flag is set or not
```
andreaslochbihler-da, line 526: Nit: `doUpdate` is a piece of code and therefore not visible to anybody. Only choices (or the corresponding exercise nodes) are visible to certain nodes. 
```suggestion
For example, if the `doUpdate` is called from a choice visible to the `owner` party, then whether the `efficient` flag is set or not
```

docs/2.9.0/docs/canton/usermanual/performance.rst
```diff
@@ -506,74 +506,104 @@ If you submit a ledger API command (which can have multiple Daml commands), then
...
 will have a huge impact on the number of views created.
 
 .. code:: daml
 
    doUpdate : Bool -> Party -> Party -> [ContractId Example] -> Update ()
-   doUpdate efficient owner obs batch = do 
-   if efficient then do 
+   doUpdate efficient owner obs batch = do
+   if efficient then do
```
andreaslochbihler-da, line 533: Mind the indentation! You must indent the body of the do further than the line that introduces the do block. So please add a few spaces here.

docs/2.9.0/docs/canton/usermanual/performance.rst
```diff
@@ -506,74 +506,104 @@ If you submit a ledger API command (which can have multiple Daml commands), then
...
          obs : Party
          -- The observer here is a choice observer. Therefore, Canton will
-         -- ship a single view with the top node of this choice if the observer of the 
-         -- choice aligns with the observer of the contract.       
-         observer obs
-         controller owner 
+         -- ship a single view with the top node of this choice if the observer of the
+         -- choice aligns with the observer of the contract. In other words, either they are the same or the new observer
+         -- is hosted by any of the contract's informee participants.
+         observer new_obs
```
andreaslochbihler-da, line 571: `new_obs` doesn't seem to be in scope here.
```suggestion
         observer obs
```

docs/2.9.0/docs/canton/usermanual/performance.rst
```diff
@@ -506,74 +506,104 @@ If you submit a ledger API command (which can have multiple Daml commands), then
...
+         controller owner
          do
          forA_ batch (`exercise` Foo)
```
andreaslochbihler-da, line 574: Indentation!

docs/2.9.0/docs/canton/usermanual/performance.rst
```diff
@@ -506,74 +506,104 @@ If you submit a ledger API command (which can have multiple Daml commands), then
...
+with the list of contracts that in turn exercises the Foo choice per stakeholder group.
 
-For example, if the below `doUpdate` is only visible to the `owner` party, then whether the `efficient` flag is set or not 
+In general, if you write batch commands, group the operations on the ledger based on the informees' participants and create
+a single choice and batch operation per informee's participants group.
+
+For example, if the `doUpdate` is only visible to the `owner` party, then whether the `efficient` flag is set or not
 will have a huge impact on the number of views created.
 
 .. code:: daml
```
andreaslochbihler-da, line 529: Given the many syntactic errors in this example, it would be IMO good to source this example from a Daml file that is actually compiled as part of CI.

docs/2.9.0/docs/canton/usermanual/performance.rst
```diff
@@ -506,74 +506,104 @@ If you submit a ledger API command (which can have multiple Daml commands), then
...
+      observer obs
+
+      -- even though the participants to be informed are a subset of the contract's informee participants
+      -- they don't get merged into a single view if this choice is called independently each time (multiple root actions)
       choice Foo : ()
-         controller owner 
-         do 
+         controller owner
+         do
          return ()
```
andreaslochbihler-da, line 555: Indentation.

docs/2.9.0/docs/canton/usermanual/performance.rst
```diff
@@ -506,74 +506,104 @@ If you submit a ledger API command (which can have multiple Daml commands), then
...
    doUpdate : Bool -> Party -> Party -> [ContractId Example] -> Update ()
-   doUpdate efficient owner obs batch = do 
-   if efficient then do 
+   doUpdate efficient owner obs batch = do
+   if efficient then do
       -- Would be even more efficient if we can reuse the BatchFoo contract
-      batcher <- create BatchFoo with owner 
-      batcher `exercise` Run with batch = batch; obs = obs    
+      batcher <- create BatchFoo with owner
+      batcher `exercise` Run with batch = batch; obs = obs
```
andreaslochbihler-da, line 536: This only works out if `obs` is the same observer on all the exercised `Example` contracts. It might be worth to mention this somewhere in the example.

docs/2.9.0/docs/canton/usermanual/performance.rst
```diff
@@ -506,74 +506,104 @@ If you submit a ledger API command (which can have multiple Daml commands), then
...
+         -- is hosted by any of the contract's informee participants.
+         observer new_obs
+         controller owner
          do
          forA_ batch (`exercise` Foo)
 
-
-As a rule, the number of views should depend on the number of stakeholder groups you have in your batch choice, 
-not the number of "batches" you process in parallel within one command.
+As a rule, the number of views should depend on the number of participant stakeholder groups
```
andreaslochbihler-da, line 576: What is a participant stakeholder group and how does it differ from the concept of informee participants?

docs/2.9.0/docs/canton/usermanual/performance.rst
```diff
@@ -506,74 +506,104 @@ If you submit a ledger API command (which can have multiple Daml commands), then
...
 The informees for the different type of transaction tree nodes are (also see :ref:`da-model-projections`):
-   * create: signatories, observers 
+   * create: signatories, observers
    * consuming exercise: signatories, observers, stakeholders of the choice (controller, choice observers)
    * nonconsuming exercise: signatories, stakeholders of the choice (controller, choice observers), but not the observers of the contract
    * fetch: signatories + actors of the fetch, which are all stakholders which are in the authorization context that the fetch executed in
    * lookupByKey: only key maintainers
 
+
+Strategy for Protocol Version 5 (and lower)
```
andreaslochbihler-da, line 590:
```suggestion
Strategy for Protocol Version 5 and Lower
```

docs/2.9.0/docs/canton/usermanual/performance.rst
```diff
@@ -506,74 +506,104 @@ If you submit a ledger API command (which can have multiple Daml commands), then
...
++++++++++++++++++++++++++++++++++++++++++++
+
+For protocol versions 5 and lower, Canton creates a view for every action node in the transaction tree if the informees
+or confirming parties change between the node and its parent. Therefore, in order to minimize the number of views,
+you should try to minimize the number of changes of informees and confirming parties.
+
+When compared with the previous decomposition strategy, this view decomposition is more granular,
+with more views created, and thus less performant, since the condition to merge an action node in a view
+is more restricted. In this case, we only consider the list of informees and any change to this list will spawn
```
andreaslochbihler-da, line 599:
```suggestion
is more restricted. In this case, we only consider the lists of informees and confirming parties and any change to these lists will spawn
```

docs/2.9.0/docs/canton/usermanual/performance.rst
```diff
@@ -506,74 +506,104 @@ If you submit a ledger API command (which can have multiple Daml commands), then
...
+a new view.
+
+To batch, the rule here is to group the operations on the ledger based solely on the list of informees.
```
andreaslochbihler-da, line 602:
```suggestion
To batch, the rule here is to group the operations on the ledger based solely on the lists of informees and confirming parties.
```

docs/2.9.0/docs/canton/usermanual/performance.rst
```diff
@@ -506,74 +506,104 @@ If you submit a ledger API command (which can have multiple Daml commands), then
...
+You can use choice observers to align informees if necessary. With the previous model as reference, the batch
+function would work if the observer of the aggregating choice is the same as the `Example` contract observer
+(`obs` == `new_obs`).
```
andreaslochbihler-da, line 605: Given that `new_obs` is unbound in the above example, it's not clear what you are aiming for. In particular, I don't see why the above example would not work with PV <= 5. Can you elaborate on the difference?

JoaoSa-da: The idea is if the `obs` in  `doUpdate efficient owner obs batch = do` is different from the one in the '_Example_' contract it won't work for PV <= 5. For PV6 it will work even if the different observer is different as long as it is hosted in an informee participant of the '_Example_' contract.

JoaoSa-da: I remove the reference to `new_obs` and try to explain this in the text.

docs/2.9.0/docs/canton/usermanual/performance.rst
JoaoSa-da: I was using them as synonyms, but in this phrase it probably creates confusion since we use _"informee participants"_ right after.

docs/2.9.0/docs/canton/usermanual/performance.rst
JoaoSa-da: I agree, I've opened an issue and will do this in a subsequent [PR](https://github.com/DACH-NY/canton/issues/20047).

docs/2.9.0/docs/canton/usermanual/performance.rst
```diff
@@ -506,74 +506,111 @@ If you submit a ledger API command (which can have multiple Daml commands), then
...
+
+Starting from protocol version 6, Canton creates a view for every action node in the transaction tree if the
+**participants that host their informees are not a subset of their parent view's informee participants**.
+A view's informee participants are all the participants that must be informed about the view.
+Therefore, to reduce the number of views, you should try to reduce either the number of times the set of informees grows
+or make sure most of your actions share the same pool of participants that must be informed. Alternatively, you can
+add informees to the action nodes close to the root (e.g., by including choice observers) so that their children nodes
+are aggregated in the initial view, since these children nodes will only target a subset of those participants.
+
+One concrete way to reduce the number of views being generated by a daml model is to write batch commands and group the
```
andreaslochbihler-da, line 520:
```suggestion
One concrete way to reduce the number of views being generated by a Daml model is to write batch commands and group the
```

docs/2.9.0/docs/canton/usermanual/performance.rst
```diff
@@ -506,74 +506,111 @@ If you submit a ledger API command (which can have multiple Daml commands), then
...
+operations on the ledger based on their informees' participants and create a single choice and batch operations per
+informee's participants group.
+
+You can also reduce the number of views being generated for the children of a node (i.e., parent-child views) when
+creating your daml model.
```
andreaslochbihler-da, line 525:
```suggestion
creating your Daml model.
```

docs/2.9.0/docs/canton/usermanual/performance.rst
```diff
@@ -506,74 +506,111 @@ If you submit a ledger API command (which can have multiple Daml commands), then
...
+With the bellow example as reference, instead of sending a batch with 20 exercise commands `Foo`,
```
andreaslochbihler-da, line 526:
```suggestion
With the below example as reference, instead of sending a batch with 20 exercise commands `Foo`,
```

docs/2.9.0/docs/canton/usermanual/performance.rst
```diff
@@ -506,74 +506,111 @@ If you submit a ledger API command (which can have multiple Daml commands), then
...
-      batcher <- create BatchFoo with owner 
-      batcher `exercise` Run with batch = batch; obs = obs    
-   else do
-      -- Canton will create one view per exercise
-      forA_ batch (`exercise` Foo)    
-   pure ()
+   doUpdate efficient owner obs batch = do
+      if efficient then do
+         -- Would be even more efficient if we can reuse the BatchFoo contract
+         batcher <- create BatchFoo with owner
```
andreaslochbihler-da, line 541: Why "more efficient"? In PV 6, the create node should end up in the same view as the calling choice calling `doUpdate` (because the owner is the only stakeholder and therefore present in the parent node). So the create doesn't change the number of views.

docs/2.9.0/docs/canton/usermanual/performance.rst
```diff
@@ -506,74 +506,111 @@ If you submit a ledger API command (which can have multiple Daml commands), then
...
+   with
+      owner : Party
+   where
+      signatory owner
 
       choice Run : ()
-         with  
+         with
          batch : [ContractId Example]
          obs : Party
```
andreaslochbihler-da, line 573: I think you want to indent these two lines too.

docs/2.9.0/docs/canton/usermanual/performance.rst
```diff
@@ -506,74 +506,111 @@ If you submit a ledger API command (which can have multiple Daml commands), then
...
-         controller owner 
+         controller owner
          do
-         forA_ batch (`exercise` Foo)
+            forA_ batch (`exercise` Foo)
 
-
-As a rule, the number of views should depend on the number of stakeholder groups you have in your batch choice, 
-not the number of "batches" you process in parallel within one command.
+As a rule, the number of views should depend on the number of informee participants
```
andreaslochbihler-da, line 583:
```suggestion
As a rule, the number of views should depend on the number of groups of informee participants
```

docs/2.9.0/docs/canton/usermanual/performance.rst
```diff
@@ -506,74 +506,111 @@ If you submit a ledger API command (which can have multiple Daml commands), then
...
+you have in your batch choice that is not a subset of the parent's contract informee participants,
```
andreaslochbihler-da, line 584:
```suggestion
you have in your batch choice,
```

docs/2.9.0/docs/canton/usermanual/performance.rst
```diff
@@ -506,74 +506,111 @@ If you submit a ledger API command (which can have multiple Daml commands), then
...
+not the number of "batches" you process in parallel within one command. A participant stakeholder group is formed
```
andreaslochbihler-da, line 585: I thought you wanted to eliminate the term "participant stakeholder group".

docs/2.9.0/docs/canton/usermanual/performance.rst
```diff
@@ -506,74 +506,111 @@ If you submit a ledger API command (which can have multiple Daml commands), then
...
 
+
+Strategy for Protocol Versions 5 and Lower
+++++++++++++++++++++++++++++++++++++++++++
+
+For protocol versions 5 and lower, Canton creates a view for every action node in the transaction tree if the informees
+or confirming parties change between the node and its parent. Therefore, in order to minimize the number of views,
+you should try to minimize the number of changes of informees and confirming parties.
+
+When compared with the previous decomposition strategy, this view decomposition is more granular,
```
andreaslochbihler-da, line 604: "previous" is ambiguous here because PV 6 was described **above**, but has been released **after** PV 5.
```suggestion
When compared with the decomposition strategy for protocol version 6, this view decomposition is more granular,
```

andreaslochbihler-da (approved):
LGTM. Left a few small comments.

docs/2.9.0/docs/canton/usermanual/performance.rst
JoaoSa-da: Makes sense. I'll remove this comment.

PR #754, opened 2024-07-15
fix docs theme build

bame-da:
I think this fixes the missing CSS file.


rgugliel-da (approved):
Works locally
Thanks for the fix

PR #766, opened 2024-07-24
Fix PQS ToC levels

filmackay:
This fixes the heading levels in the PQS docs, so it displays appropriately in the ToC.


yves-da (approved):
LGTM, thank you.

(I assume that you've tested it locally and verified the fix.)

dasormeter (approved):

cathyjung-da (approved):

PR #791, opened 2024-09-02
Update daml 2.10's upgrade documentation with latest changes

paulbrauner-da:
- Update the doc to mention LF 1.17 instead of 1.16
- Update the doc with the latest restrictions wrt to interfaces

### DIFF
--- a/docs/2.10.0/docs/concepts/ledger-model/upgrade-model-in-depth.rst
+++ b/docs/2.10.0/docs/concepts/ledger-model/upgrade-model-in-depth.rst
@@ -1080,7 +1080,7 @@ whole.
 **Examples**
 
 In these examples we assume the existence of packages ``q-1.0.0`` and
-``q-2.0.0`` with LF version 1.16, and that the latter is a valid upgrade of
+``q-2.0.0`` with LF version 1.17, and that the latter is a valid upgrade of
 the former.
 
 .. list-table::
--- a/docs/2.10.0/docs/concepts/ledger-model/upgrade-model-in-depth.rst
+++ b/docs/2.10.0/docs/concepts/ledger-model/upgrade-model-in-depth.rst
@@ -1311,16 +1311,13 @@ package of their own that defines no template.
 Interface Instances
 ~~~~~~~~~~~~~~~~~~~
 
-The instances of an interface in the new version of a package must form a
-superset of the instances of that interface in the prior version of that 
-package. In other words, it is valid to add new interface instances but
-deleting an interface instance leads to a validation error.
-
 Interface instances may be upgraded. Note however that the type signature of 
 their methods and view cannot change between two versions of an instance since
 they are fixed by the interface definition, which is non-upgradable. Hence,
 the only thing that can change between two versions of an instance is the
 bodies of its methods and view.
+
+Adding or deleting an interface leads to a validation error.
 
 **Examples**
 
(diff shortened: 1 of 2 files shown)


docs/2.10.0/docs/concepts/ledger-model/upgrade-model-in-depth.rst
```diff
@@ -2023,3 +2033,123 @@ have been written using the template on the left with ``n = 0``.
...
+.. code:: daml
+
+    data IView = IView { i : Int }
+  
+    interface I where
+      viewtype IView
+      m : Int
+ 
+Then below the template on the right is a valid upgrade of the template on the
+left.
```
carrielaben-da, line 2059:
```suggestion
In that case, the template on the right below is a valid upgrade of the template on the
left.
```

carrielaben-da (approved):
One small suggestion for clarity.

docs/2.10.0/docs/upgrade/smart-contract-upgrades.rst
```diff
@@ -163,7 +163,7 @@ To prevent unexpected behavior, this feature enforces a unique package name and
 uploaded to a participant node.
 This closes a loophole where the participant node allowed multiple DARs with
 the same package name and version. For backward compatibility, this
-restriction only applies for packages compiled with LF >= 1.16. If LF <
+restriction only applies for packages compiled with LF >= 1.17. If LF <
 1.15 is used, there can be several packages with the same name and
```
dylant-da, line 167: The following line probably needs updating to be "if LF < 1.16", or a note needs to be made of the behaviour in 1.16

dylant-da (approved):
Aside from comments, looks good - are we burning 1.16 completely? I didn't realize

docs/2.10.0/docs/upgrade/smart-contract-upgrades.rst
paulbrauner-da: As discussed offline, everything is as if 1.16 had never existed, so it's fine for its behavior to be unspecified.

samuel-williams-da:
I believe these docs also exist in 3.x, do they need the same changes?

docs/2.10.0/docs/upgrade/smart-contract-upgrades.rst
```diff
@@ -1916,11 +1932,11 @@ JSON API Server
...
 
 Template IDs may still be used with a package ID, however,
-for packages built as LF 1.16 or greater, the package may also be
+for packages built as LF 1.17 or greater, the package may also be
 identified by name. That is to say, for upgradable packages a template ID can have
 the form ``#<package-name>:<module-name>:<template-name>``, and this is
 resolved to corresponding templates from all packages which share this
-name, and are built at 1.16 or above. For packages built at LF 1.15 or
+name, and are built at 1.17 or above. For packages built at LF 1.15 or
 lower, the templates are not identifiable via a package name, and a
```
remyhaemmerle-da, line 1940:
```suggestion
name, and are built at 1.17 or above. For packages built at LF 1.15,
the templates are not identifiable via a package name, and a
```

remyhaemmerle-da (approved):
Nice

docs/2.10.0/docs/upgrade/smart-contract-upgrades.rst
```diff
@@ -1630,37 +1626,57 @@ DARs. They should both succeed again:
...
   DAR upload succeeded.
   > cd ../../v2/my-pkg
   > daml build
-  > daml sandbox upload-dar --port 6865
+  > daml ledger upload-dar --port 6865
+  ...
+  Uploading .daml/dist/my-pkg-2.0.0.dar to localhost:6865
+  upload-dar did not succeed: ... Implementation of interface ...:MyIface:HasValue by template IOU appears in package that is being upgraded, but does not appear in this package.
+
+Packages with new versions cannot add an interface instance to an existing
```
yangzai, line 1661: Hi @paulbrauner-da , under "Best Practices" > "Remove Retroactive Instances", it still states that "retroactive interface instances should be moved to newer versions of templates, such that changes to the instance warrants a new version of the template".

Doesn't it conflict with the new restriction here that "packages with new versions cannot add an interface instance to an existing template"?

paulbrauner-da: Thank you for pointing this out, we'll fix the inconsistency.

PR #792, opened 2024-09-05
Remove "N times" from the Postgres tuning guidance

wallacekelly-da:
In the guidance about tuning Postgres performance, the following phrase appears:

```
Also increase the following variables N times beyond their default
```

That has been (incorrectly) understood to mean there is some specific value of `N` to target. Later, the paragraph explains that one must "Monitor your system during load testing and tune the parameters accordingly to your use case."

I propose removing `N times` to avoid misunderstanding.


carrielaben-da (approved):
LGTM

PR #799, opened 2024-09-19
Updated SC upgrading limitations

daravep:
Before this change, we still had interfaces as being upgradable. Now, we are explicit about that interfaces and exceptions are static.


carrielaben-da (approved):
LGTM

Issue #758, opened 2024-07-16
Website `theme` only changes are ignored for deployment


yves-da:
### Problem
This PR #754 changed only content in the `theme` directory, but that change did not get pushed to docs.daml.com.

### Workaround
The `theme` changes were then pushed to docs.daml.com by means of small documentation changes, see #755 and #757.

### Analysis
On CI, the `push_docs` job didn't deploy to docs.daml.com with the following output:
```
2.10.0 is up-to-date, nothing to do.
2.3.20 is up-to-date, nothing to do.
2.7.9 is up-to-date, nothing to do.
2.8.9 is up-to-date, nothing to do.
2.9.0 is up-to-date, nothing to do.
3.1 is up-to-date, nothing to do.
draft is up-to-date, nothing to do.
root is up-to-date (2.8.9, a4fb2f8e136bc6811883dd3d8c3f23251a8268ba).
snapshots.json is up-to-date.
versions.json is up-to-date.
/downloads up-to-date, nothing to do
``` 

The `push_docs` job has got a change detection logic which does not trigger on `theme` directory only changes.
https://github.com/digital-asset/docs.daml.com/blob/main/.circleci/config.yml#L163

### Solution idea
Add the `theme` to the sha logic, that is don't take the latest commit that touched a folder, but the latest that touched a folder *or* `theme`.

Issue #722, opened 2024-06-17
Misaligned viewport when using cross-referencing using roles


yves-da:
### Problem
When using [links to explicit targets](https://www.sphinx-doc.org/en/master/usage/referencing.html#cross-referencing-arbitrary-locations), then the navigation does not align the viewport properly.

Example:
* Click on a link which uses cross-referencing, for example _versioning guide_.
<img width="1280" alt="Screenshot 2024-06-17 at 08 22 01" src="https://github.com/digital-asset/docs.daml.com/assets/119872801/236d246d-ba9f-4194-9d90-b1af6508d919">

* We navigate to the _versioning guide_ page, but our link target is hidden below the navigation bar:
<img width="1279" alt="Screenshot 2024-06-17 at 08 22 57" src="https://github.com/digital-asset/docs.daml.com/assets/119872801/c90129f3-cc4b-4e12-b120-3cf08e24f731">

* We need to scroll up so that we see the intended link location on the page.
<img width="1279" alt="Screenshot 2024-06-17 at 08 24 04" src="https://github.com/digital-asset/docs.daml.com/assets/119872801/4c8f1520-a7d9-4222-98f7-a97860498c22">

### Notes
* Implicit links (like `` `Section title`_ ``) work but can only be used on the page itself. And if someone changes the `Section title` then those links break silently. Cross-referencing links emit a compilation error when they break.

yves-da:
Thanks to @wallacekelly-da efforts, this has been fixed with:
* #813

Only caveat: Above changes only becomes effective when there's an update to the docs RST files (see #758).

PR #813, opened 2024-10-15
Fix scroll position for cross references links

wallacekelly-da:
You may have noticed a very annoying behavior when clicking on links within the Ledger API reference docs. For example, if you click on the link for the [GetActiveContractsResponse message](https://docs.daml.com/app-dev/grpc/proto-docs.html#com-daml-ledger-api-v1-getactivecontractsresponse), the heading for that message is _behind_ the navbar. It is very confusing for new users.

This PR makes a small CSS tweak to address that issue. Using jQuery, it dynamically adds the `section` CSS class to the `span`s which serve as the anchor for internal links.

I did look at trying to add that CSS class to the RST source file, but without success. I have also [asked on StackOverflow](https://stackoverflow.com/questions/79091778/add-css-class-to-sphinx-generated-anchor-span).

Attached are videos showing _before_ and _after_.

https://github.com/user-attachments/assets/a3963573-f321-499f-8dd7-3a13adaa8987

https://github.com/user-attachments/assets/66f9c6b1-a31c-4c1c-8e4b-73703a0550fd

### DIFF
--- a/theme/js/theme.js
+++ b/theme/js/theme.js
@@ -607,6 +607,11 @@ $(document).ready(function () {
             });
     }
     setHighlighterWidth();
+    
+    // Assign the CSS class "section" for Ledger API reference doc spans.
+    // This fixes a problem with clicking on internal links in the Ledger API reference docs.
+    // The scroll position, after clicking a link, was behind the nav bar.
+    $("span[id^=com-daml-ledger-api-v]").addClass("section");
 });
 
 $(window).resize(function () {
--- a/theme/js/theme.js
+++ b/theme/js/theme.js
@@ -608,10 +608,11 @@ $(document).ready(function () {
     }
     setHighlighterWidth();
     
-    // Assign the CSS class "section" for Ledger API reference doc spans.
-    // This fixes a problem with clicking on internal links in the Ledger API reference docs.
+    // Assign the CSS class "section" to any span that is a direct child of a div of class section.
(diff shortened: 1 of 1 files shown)


carrielaben-da (approved):
LGTM

yves-da:
Oh nice!

This is however a general problem on docs.daml.com, see
* https://github.com/digital-asset/docs.daml.com/issues/722

Not sure how easy it would be to apply this fix generally 🤔

wallacekelly-da:
> Not sure how easy it would be to apply this fix generally

@yves-da, with 91f6986ab81014dfcbc9728432e5e59ef39dac26 I changed the jQuery selector from `span[id^=com-daml-ledger-api-v]` to `div.section > span`.

* The former selects spans whose ids start with `com-daml-ledger-api-v`.
* The latter selects spans that are direct children of a `div` with the class `section`. This is the pattern that Sphinx uses, as described [here](https://stackoverflow.com/q/79091778/167920).

With the new selector, the issue seems to be addressed site wide.

yves-da:
@wallacekelly-da thanks a lot for the fix.

> With the new selector, the issue seems to be addressed site wide.
I wasn't able to verify that for myself. However, since this is `theme`-only change CI didn't update docs.daml.com because of https://github.com/digital-asset/docs.daml.com/issues/758

yves-da:
And indeed, after docs/2.10.0 got updated from another PR, the improvement materialized for that version. Very nice 😆

PR #825, opened 2024-11-04
Remove duplicated lualatex lines from the build command

pawelulewicz-da:
Remove the duplicated lualatex lines from the build command. It saves about 1,5 minutes locally.


yves-da (approved):
LGTM, thanks!

I doubt that removing this duplication has any adverse effect. – Still, it'd be good to quickly check the resulting PDF for any unwanted changes. I suspect that `lualatex` would only affect the rendering of formulas; so look for that. Or maybe you can just compare the resulting PDFs for any change.

pawelulewicz-da:
@yves-da Actually it breaks the ToC. Thanks for being cautious. 

Closing this PR then as there doesn't seem to be an easy fix.

Screenshot 2024-11-04 at 10 28 33

PR #826, opened 2024-11-04
Add comments for double lualatex commands

pawelulewicz-da:
Originally, I planned to remove the duplicated lines, but it turned out they were actually needed: https://github.com/digital-asset/docs.daml.com/pull/825.

The second best thing to do is to add comments with explanations.


yves-da (approved):
Thanks for the clarifying comment!

pawelulewicz-da:
As discussed, there's a problem with merging from forks.

New PR: https://github.com/digital-asset/docs.daml.com/pull/827

PR #833, opened 2024-11-12
Replace daml code snippet with the actual content of the code block

JoaoSa-da:


yves-da (approved):
Thanks @JoaoSa-da

yves-da:
@JoaoSa-da Maybe it's better to apply the same approach also for the 3.1 version.

dylant-da (approved):

PR #836, opened 2024-11-18
Add a script for listing outdated files

pawelulewicz-da:
@carrielaben-da suggested in a discussion:

> A nice-to-have for me: some mechanism that can flag pages that have not been updated within the past year as candidates for audit. Occasionally we have outdated material that persists because there hasn't been a PR that explicitly removes it (e.g. the old docs on supporting Ethereum and Hyperledger Fabric.)

Basic usage:

```
./bin/list-outdated-files
```

Customized usage:

```
./bin/list-outdated-files -l 3.months -d docs/3.1/docs -e json -o gitlog
```

Help:

```
./bin/list-outdated-files -h
```

### DIFF
--- a/bin/list-outdated-files
+++ b/bin/list-outdated-files
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
 
+# Use the following bash script for searching pages that have not been updated for a long time (default: 1 year) as candidates for audit.
+# Occasionally we have outdated material that persists because there hasn't been a PR that explicitly removes it.
+
 set -euo pipefail
 
 directory=docs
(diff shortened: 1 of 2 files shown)


bin/list-outdated-files
yves-da, line 1: The scope of this `bin` directory seems to be about docs.daml.com build concern rather then about the documentation concern. It may or may not belong here; each docs version has also its own bin directory `docs/<version>/bin`.

I think it'd be helpful to document the existence of this script in a `README.md`. There's one on the top-level or then also per docs version.

yves-da (approved):
LGTM, thanks!

The scripts looks reasonable, but I haven't checked it in detail. It may be helpful to comment what it does for lines that are not immediately obvious. I'd also add a desciption to it that explains the motivation behind this script.

bin/list-outdated-files
pawelulewicz-da: Thanks, I updated `README.md` as you requuested.

As for the file location, I chose to keep it where it was, because it's useful for all versions and I didn't want to copy-paste it to each version.

pawelulewicz-da: I also refactored the code, so it uses a separate function for action, and added some comments. Now it looks like this:

```
list_unchanged_files | filter_directory | filter_extensions | produce_output
```

PR #858, opened 2024-12-10
Fix upgrade table formatting

samuel-williams-da:
Used to look like this:
image

Now looks like this:
image


carrielaben-da (approved):
Thanks!

PR #865, opened 2024-12-12
document that the upgrade check allows for removing observers that are also signatories

paulbrauner-da:
Addresses UPD-11


carrielaben-da (approved):
All good!

samuel-williams-da (approved):
Content looks good from a technical side, thanks Paul!

PR #873, opened 2024-12-18
UPD-1 and UPD-2

carlpulley-da:
- `Migrating to SCU` section discusses conflicts that might arise due to use of SCU and package names
- latter sections (e.g. Overview/JSON API Server and PQS) discuss relationship between package-id and (package-name, package-version)
- `Migrating to SCU` section discusses that package-names can only be associated with single package-versions


docs/2.10.0/docs/upgrade/smart-contract-upgrades.rst
```diff
@@ -98,6 +98,16 @@ error-prone migrations, without downtime, without requiring any
...
 
+With SCU, different versions of the _same_ package are now grouped under a common package-name.
+This allows template/interface IDs to now be unambiguously referenced using either:
+
+-  the package ID
+
+-  or by using the package-name (which is resolved by a Canton participant to a unambiguous package ID).
+
+When defining a Daml project, the project-name field of the project's `daml.yaml` is now used to
+specify the SCU project-name.
```
samuel-williams-da, line 109:
```suggestion
When defining a Daml package, the ``name`` field of the package's ``daml.yaml`` is now used to
specify the SCU package-name.
```

samuel-williams-da: Be careful to use `package` over `project`, as we like to use project now to describe a set of packages.
Also docs.daml.com requires double backticks for code sections

docs/2.10.0/docs/upgrade/smart-contract-upgrades.rst
```diff
@@ -98,6 +98,16 @@ error-prone migrations, without downtime, without requiring any
 additional network traffic, and without any extensive communication with
 downstream users.
 
+With SCU, different versions of the _same_ package are now grouped under a common package-name.
+This allows template/interface IDs to now be unambiguously referenced using either:
+
+-  the package ID
+
+-  or by using the package-name (which is resolved by a Canton participant to a unambiguous package ID).
```
paulbrauner-da, line 106: Since we want to promote the best practice of providing package preferences in ledger API request, and also because it makes it sound less magical, I would rather communicate that a package is uniquely determined by a package ID or by a package name + a package preference (which is a set of preferred package IDs in which no two package IDs have the same name, i.e. an implicit mapping from package names to package IDs). This is an overview so this formulation is way too verbose. The point I'd like to get across is that the user has control over how the participant resolves names and more importantly, we recommend using that control.

docs/2.10.0/docs/upgrade/smart-contract-upgrades.rst
```diff
@@ -98,6 +98,16 @@ error-prone migrations, without downtime, without requiring any
 additional network traffic, and without any extensive communication with
 downstream users.
 
+With SCU, different versions of the _same_ package are now grouped under a common package-name.
```
paulbrauner-da, line 101: It sounds a bit as if you're saying "these packages are different versions of the same package so they must share a name", while the way I'm thinking about it is: "giving the same name to two packages indicates your intent for them to be considered different versions of the same package". Granted, the difference is subtle and my formulation is awkward, I'm not sure how to best express it.

samuel-williams-da (commented):
Side note, In Carrie's last review, she replaced `package-name` with `package name`
dropping the dash. You'll likely want to do that here as well, to match

docs/2.10.0/docs/upgrade/smart-contract-upgrades.rst
carlpulley-da: Thanks for this. You've made me realise the importance of the following 2 properties:
- package versions are totally ordered
- package preference maps relate package names to package versions using an injective/1:1 relationship.

docs/2.10.0/docs/upgrade/smart-contract-upgrades.rst
carlpulley-da: There is a latter section of the document (Dynamic Package Resolution in Ledger API Queries) that better describes this process. As a result I've removed this text from the overview.

docs/2.10.0/docs/upgrade/smart-contract-upgrades.rst
carlpulley-da: Refactored text to keep it simple

docs/2.10.0/docs/upgrade/smart-contract-upgrades.rst
```diff
@@ -98,6 +98,20 @@ error-prone migrations, without downtime, without requiring any
 additional network traffic, and without any extensive communication with
 downstream users.
 
+With SCU, packages may now be referenced by either:
+
+-  the package ID
+
+-  or by using a package name.
```
carrielaben-da, line 105:
```suggestion
-  the package name.
```

carrielaben-da (approved):
Small changes

PR #878, opened 2024-12-18
Document UPD-24 and UPD-28

dylant-da:


docs/2.10.0/docs/upgrade/smart-contract-upgrades.rst
```diff
@@ -2008,6 +2011,152 @@ against those packages and its datatypes are not automatically converted.
...
+  import qualified Daml.Script
+
+  data MyData = MyData
+    { field1 : Daml.Script.PartyIdHint
+    }
+
+Because ``MyData`` is a serializable datatype, any changes to it must be valid
+upgrade changes (e.g. adding a field) for the ``main`` package itself to be
+validly upgraded. If SDK 2.10.X were to introduce a change to
+``daml-script-lts`` that is not a valid upgrade of the ``daml-script-lts`` in
```
samuel-williams-da, line 2110: It might be worth mentioning that its very very unlikely that we'll do this, for peace of mind of those who have already fallen prey, but up to you. It is somewhat counter to our point

docs/2.10.0/docs/upgrade/smart-contract-upgrades.rst
```diff
@@ -2008,6 +2011,152 @@ against those packages and its datatypes are not automatically converted.
...
+    { field1 : Text
+    }
+
+  myFunction : Daml.Script.PartyIdHint -> Bool
+  myFunction _ = True
+
+In this case, when changing SDK from 2.10.0 to 2.10.X, the typechecker will
+ignore the change of ``Daml.Script.PartyIdHint`` in ``myFunction``, because it
+is not in a serializable position. This means ``daml-script-lts`` can be kept
+even though when it is not a valid upgrade from one version to the next.
```
samuel-williams-da, line 2158: Note that the upload check would still warn in this case though. We warn if daml-script is uploaded at all.
I realise we discussed this before in the context of depending on 1.15 packages.
Its okay to depend on 1.15 if not in a serializable position
Its less okay to depend on daml-script, even if not serializable, if its uploaded.
We should probably just add a caveat about this. It might need rethinking to be a nicer experience.

samuel-williams-da: I would say that having a 1.15 type in your template is worse than having a daml-script type though, since we are being careful with script. Up to you if you want to convey that

docs/2.10.0/docs/upgrade/smart-contract-upgrades.rst
dylant-da: Updated.

docs/2.10.0/docs/upgrade/smart-contract-upgrades.rst
```diff
@@ -2008,6 +2011,159 @@ against those packages and its datatypes are not automatically converted.
...
+We recommend only depending on Daml Script packages such as ``daml-script-lts``
+in dedicated packages for running tests written in Daml Script. These packages
+should not be part of your model and should not be uploaded to the ledger.
+
+Daml Script packages are not guaranteed to be upgradeable across SDK versions.
+For this reason, if you depend on a Daml Script datatype in a serializable
+position (e.g. the field of a template), you may find your package will rely on
+a Daml Script package in a way that can neither be removed nor upgraded to the
+next SDK version. At this point, your package and any of its SCU upgrades would
+be stuck on that SDK version.
```
carrielaben-da, line 2080:
```suggestion
Daml Script packages are not guaranteed to be upgradeable across SDK versions.
If you depend on a Daml Script datatype in a serializable
position (e.g. the field of a template), your package may rely on
a Daml Script package in a way that can neither be removed nor upgraded to the
next SDK version. Your package and any of its SCU upgrades would
be stuck on that SDK version.
```

docs/2.10.0/docs/upgrade/smart-contract-upgrades.rst
```diff
@@ -2008,6 +2011,159 @@ against those packages and its datatypes are not automatically converted.
...
+  data MyData = MyData
+    { field1 : Daml.Script.PartyIdHint
+    }
+
+Because ``MyData`` is a serializable datatype, any changes to it must be valid
+upgrade changes (e.g. adding a field) for the ``main`` package itself to be
+validly upgraded. If SDK 2.10.X were to introduce a change to
+``daml-script-lts`` that is not a valid upgrade of the ``daml-script-lts`` in
+SDK 2.10.0, then ``field1`` will not be upgradeable to the next version of the
+SDK, nor can the field be dropped because the field is used in SCU checks.
```
carrielaben-da, line 2112:
```suggestion
Because ``MyData`` is a serializable datatype, any changes to it must be valid
upgrade changes (e.g. adding a field) for the ``main`` package itself to be upgraded. If SDK 2.10.X introduces a change to
``daml-script-lts`` that is not a valid upgrade of the ``daml-script-lts`` in
SDK 2.10.0, then ``field1`` is not upgradeable to the next version of the
SDK, nor can the field be dropped because the field is used in SCU checks.
```

docs/2.10.0/docs/upgrade/smart-contract-upgrades.rst
```diff
@@ -2008,6 +2011,159 @@ against those packages and its datatypes are not automatically converted.
...
+
+**Note**: While it is unlikely that an SDK update will break ``daml-script-lts``
+for upgrades, we still strongly recommend against it. In Daml 3.x, Canton may
+disallow uploading ``daml-script-lts`` entirely.
```
carrielaben-da, line 2116:
```suggestion
**Note**: While it is unlikely that an SDK update breaks ``daml-script-lts``
for upgrades, we still strongly recommend against it. In Daml 3.x, Canton may
disallow uploading ``daml-script-lts`` entirely.
```

docs/2.10.0/docs/upgrade/smart-contract-upgrades.rst
```diff
@@ -2008,6 +2011,159 @@ against those packages and its datatypes are not automatically converted.
...
+.. code:: bash
+
+  > daml build
+  ...
+  warning while type checking data type Main.MyData:
+    This package depends on a datatype <package ID>:Daml.Script.Stable:PartyIdHint from <package ID> (daml-script-lts-stable, 0.0.0) with LF version 1.17.
+    
+    It is not recommended that >= LF1.17 packages use datatypes from Daml Script, because those datatypes will not be upgradeable.
+    Upgrade this warning to an error -Werror=upgrade-serialized-daml-script
+    Disable this warning entirely with -Wno-upgrade-serialized-daml-script
```
carrielaben-da, line 2136:
```suggestion
    We do not recommend using datatypes from Daml Script in >= LF1.17 packages, because those datatypes are not upgradeable.
    Upgrade this warning to an error -Werror=upgrade-serialized-daml-script
    Remove this warning with -Wno-upgrade-serialized-daml-script
```

docs/2.10.0/docs/upgrade/smart-contract-upgrades.rst
```diff
@@ -2008,6 +2011,159 @@ against those packages and its datatypes are not automatically converted.
...
+
+**Note:** For added safety, you may upgrade these warnings to errors with
+``-Werror=upgrade-serialized-daml-script``. We recommend against disabling these
+warnings with ``-Wno-upgrade-serialized-daml-script``.
```
carrielaben-da, line 2140:
```suggestion
**Note:** For added safety, you may upgrade these warnings to errors with
``-Werror=upgrade-serialized-daml-script``. We recommend against removing these
warnings with ``-Wno-upgrade-serialized-daml-script``.
```
Again, seems redundant.

docs/2.10.0/docs/upgrade/smart-contract-upgrades.rst
```diff
@@ -2008,6 +2011,159 @@ against those packages and its datatypes are not automatically converted.
...
+
+If instead ``main`` depended on a datatype in a non-serializable position such
+as the type signature of a function, ``main`` could still be upgraded without
+breaking any SCU restrictions. For example:
```
carrielaben-da, line 2144:
```suggestion
If instead ``main`` depends on a datatype in a non-serializable position such
as the type signature of a function, ``main`` can still be upgraded without
breaking any SCU restrictions. For example:
```

docs/2.10.0/docs/upgrade/smart-contract-upgrades.rst
```diff
@@ -2008,6 +2011,159 @@ against those packages and its datatypes are not automatically converted.
...
+    { field1 : Text
+    }
+
+  myFunction : Daml.Script.PartyIdHint -> Bool
+  myFunction _ = True
+
+In this case, when changing SDK from 2.10.0 to 2.10.X, the typechecker will
+ignore the change of ``Daml.Script.PartyIdHint`` in ``myFunction``, because it
+is not in a serializable position. This means ``daml-script-lts`` can be kept
+even though when it is not a valid upgrade from one version to the next.
```
carrielaben-da, line 2162:
```suggestion
In this case, when changing SDK from 2.10.0 to 2.10.X, the typechecker
ignores the change to ``Daml.Script.PartyIdHint`` in ``myFunction``, because it
is not in a serializable position. This means ``daml-script-lts`` can be kept
even when it is not a valid upgrade from one version to the next.
```

carrielaben-da (approved):
Small changes.

docs/2.10.0/docs/upgrade/smart-contract-upgrades.rst
```diff
@@ -2008,6 +2008,157 @@ against those packages and its datatypes are not automatically converted.
...
+    Upgrade this warning to an error -Werror=upgrade-serialized-daml-script
+    Remove this warning with -Wno-upgrade-serialized-daml-script
+
+**Note:** For added safety, you may upgrade these warnings to errors with
+``-Werror=upgrade-serialized-daml-script``. We recommend against removing these
+warnings with ``-Wno-upgrade-serialized-daml-script``.
+
+If instead ``main`` depends on a datatype in a non-serializable position such
+as the type signature of a function, ``main`` can still be upgraded without
+breaking any SCU restrictions. For example:
```
paulbrauner-da, line 2139: I'm wondering if we shouldn't drop that part. It may give a false sense of security to the users. There's still the problem that if you upload two different versions of daml-script-lts to the ledger, they may not be compatible. Then again, if we ever break that, we should ear our own dogfood and call it daml-script-lts-v2.

paulbrauner-da (approved):
This is great!

docs/2.10.0/docs/upgrade/smart-contract-upgrades.rst
dylant-da: I've been playing with rewriting that section, and couldn't find anything better than a the bolded Note approach that we already have. I don't want to unmention it - I'd rather customers be aware of a workaround, along with its caveats, than that they discover it on their own and think it's a solution with no drawbacks.

PR #891, opened 2025-01-06
Deprecate Intel Mac

bame-da:


hrischuk-da (approved):

docs/2.10.0/docs/ops/requirements.rst
```diff
@@ -8,8 +8,10 @@ System Requirements
 
 Unless otherwise stated, all Daml runtime components require the following:
 
-1. For development, an x86-compatible system running a modern Linux, Windows,
-   or MacOS operating system
+1. For development, one of
+  1. x86-compatible systems running a modern Linux or Windows operating system.
+  2. ARM-based Macintosh systems running MacOS 11 or above.
+  3. Deprecated: Intel-based Macintosh systems.
```
daravep, line 14: `Deprecated` is relatively unspecific. It could mean that it's supported for 2.10 but not 2.11 or not 3.x etc. So I would be explicit and say "Unsupported".

mziolekda: I agree

bame-da: I'm with Curtis here. Unsupported suggests that if you file a support ticket on Intel Mac on 2.10 we'll just say "no". That's not what will happen. In reality, if we get a support ticket on Intel Mac and we can no longer reasonably fix things for Intel Mac because neither Apple nor other vendors/dependency are available anymore, the solution we'll offer is to move to new hardware.

docs/2.10.0/docs/ops/requirements.rst
```diff
@@ -8,8 +8,10 @@ System Requirements
 
 Unless otherwise stated, all Daml runtime components require the following:
 
-1. For development, an x86-compatible system running a modern Linux, Windows,
-   or MacOS operating system
+1. For development, one of
```
carrielaben-da, line 11:
```suggestion
1. For development, one of:
```

docs/2.10.0/docs/ops/requirements.rst
carrielaben-da, line 14:
```suggestion
  * x86-compatible systems running a modern Linux or Windows operating system
  * ARM-based Macintosh systems running MacOS 11 or above
  * Deprecated: Intel-based Macintosh systems 
```

carrielaben-da (approved):

PR #895, opened 2025-01-10
Remove Explicit Template Versions from upgrading docs

tudor-da:
Decided together with @meiersi-da and @hrischuk-da to remove as it's not targeting general enough use cases and might lead to confusion. CC @remyhaemmerle-da


meiersi-da (approved):
Thx a lot @tudor-da :pray

PR #903, opened 2025-01-24
Release 2.9.6 (#898)

yves-da:


docs/2.9.6/versions.json
yves-da, line 1: @johan-da Is the Daml Finance dependency version correct for the 2.9.6 documentation?

rgugliel-da (approved):

docs/2.9.6/versions.json
yves-da: Checked with Johan; yes, it's fine.

carrielaben-da (approved):
Thanks!

PR #912, opened 2025-02-12
Intro added to provide context and extra resource

cathyjung-da:
I noticed that I should've provided the link to a relevant resource, which covers a different aspect of the main topic.


carrielaben-da (approved):
LGTM

